PR #28, opened 2019-03-13
Updating Public User module with changes from NLW/IDA

stephenwf:
* Adds new OAuth authentication to allow configured 3rd party
  applications to authenticate using Madocs public user module.
* Fixes Annotation Studio 500 error that can sometimes occur when
  partOf field is a object {id, type} instead of a string.
* Fixed bug where automated registrations would not work.


repos/public-user/config/services.config.php
```diff
@@ -22,6 +25,27 @@
                 );
             },
 
+            AuthSettings::class => function (ContainerInterface $container) {
+                // @todo hook up to actual settings.
+                return new AuthSettings(
+                    AuthSettings::EXAMPLE_SETTINGS['clients'],
+                    AuthSettings::EXAMPLE_SETTINGS['scopes']
```
stephenwf: I don't think the scope allows for this to be editable in a UI, but we could allow this to be part of the `application.config.php` and source it from there. I will hook that up before merging.

PR #30, opened 2019-03-14
Pagination

stephenwf:


Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN dnf update -y && \
         php-opcache \
         php-json \
         php-pear \
+        php-apcu \
```
stephenwf: Omeka checks for APCU and uses this as a cache in Doctrine, added to help production caching.

config/composer.json
```diff
@@ -2,6 +2,7 @@
   "name": "digirati/other-vendors",
   "license": "MIT",
   "require": {
+    "doctrine/orm": "^2.5",
```
stephenwf: This `composer.json` is for development only, to get the correct dependencies available

repos/iiif-storage/src/Controller/CollectionController.php
```diff
@@ -44,6 +44,9 @@ public function __construct(CollectionRepository $repo, CollectionBuilder $build
         $this->settingsHelper = $settingsHelper;
     }
 
+    // @todo move this configuration.
+    const PER_PAGE = 24;
```
stephenwf: Added 24 for its divisibility. Not sure if we do want this to be completely configurable, maybe a multiple choice. The settings is available in this class so can be added in quickly.

repos/iiif-storage/src/JsonBuilder/CollectionBuilder.php
```diff
@@ -28,31 +31,37 @@ class CollectionBuilder
...
-                return $acc;
-            }, []
-        );
         // Build collection.
-        $collection = ResourceFactory::createCollection($json, function ($url) use ($manifestIndex, $originalIds) {
-            return $this->manifestBuilder->build($manifestIndex[$url], $originalIds);
+        $collection = ResourceFactory::createCollection($json, function (string $url, Manifest $manifest) use ($originalIds) {
+            $metadata = $manifest->getSource();
+            if (!isset($metadata['o:id'])) {
+                return (array)$manifest->getSource();
```
stephenwf: This trick here short-cuts loading in ALL of the manifests on the collection. Allows us to lazy-load without issue.

repos/iiif-storage/src/Repository/CollectionRepository.php
```diff
@@ -59,6 +84,23 @@ public function getDefaultQuery()
...
+        $statement->execute();
+
+        $indexMap = [];
+        foreach ($statement->fetchAll() as $manifest) {
+            if ($manifest['uri']) {
+                $indexMap[$manifest['uri']] = $manifest['omeka_id'];
+            }
+        }
+        return $indexMap;
+    }
```
stephenwf: This was written but not used in the end. The SQL up at the top is the quickest way to get a list of manifests on a collection, with omeka ids and labels.

Dockerfile
garyttierney: What are we using APCu to cache? It's pretty old these days.

stephenwf: Used for Doctrine cache by Omeka: https://github.com/omeka/omeka-s/blob/b9c9a9790fd5e539fe6225b92fe546d1d872975c/application/src/Service/EntityManagerFactory.php#L54

garyttierney: Ah, just the annotation metadata cache.

stephenwf: yeh, still no query cache as far as I can tell

PR #34, opened 2019-03-15
Split docker into base build and full build

stephenwf:


docker/madoc-omeka-s/usr/local/bin/madoc-installer
```diff
@@ -0,0 +1,17 @@
...
+patch-composer
+composer update --lock --optimize-autoloader --no-dev --prefer-source --no-interaction
+
+## Clear caches.
+composer clearcache
+rm -Rf /srv/omeka/.local /srv/omeka/.config /srv/omeka/.composer
+
+## Ensure files exists.
+mkdir -p /srv/omeka/files
+chmod 777 /srv/omeka/files
```
garyttierney: We should really be getting these permissions correct.

docker/madoc-omeka-s/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+    chown -R www-data:www-data /srv/omeka /var/www
+
+# Composer cache warming.
+# Can't enable this until we remove the `repositories`
+#RUN curl -o /srv/omeka/composer.json -sSL "https://raw.githubusercontent.com/digirati-co-uk/omeka-s/bugfix/7.2-rebase/composer.json" && \
+#    curl -o /srv/omeka/composer.lock -sSL "https://raw.githubusercontent.com/digirati-co-uk/omeka-s/bugfix/7.2-rebase/composer.lock" && \
+#    composer update --working-dir=/srv/omeka --lock --optimize-autoloader --no-dev --prefer-source --no-interaction
+
+# Download our fork.
+ADD --chown=www-data:www-data https://github.com/digirati-co-uk/omeka-s/archive/bugfix/7.2-rebase.tar.gz /srv/omeka/
```
garyttierney: Can we add a SHA-sum check for the tarball?

docker/madoc-omeka-s/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+# Custom overrides to Omeka core.
+ADD --chown=www-data:www-data srv/omeka/application/view/error/index.phtml /srv/omeka/application/view/error/index.phtml
+
+# Extended base configuration
+ADD --chown=www-data:www-data srv/omeka/application/config/*.config.php /srv/omeka/application/config/
+ADD --chown=www-data:www-data srv/omeka/config/*  /srv/omeka/config/
+
+# Binaries
+ADD --chown=www-data:www-data usr/local/bin/patch-composer /usr/local/bin/
```
garyttierney: Might've overlooked something, but this is named `patch-composer.php` in other places, the extension is dropped here.

stephenwf: Made a copy of the original patch-composer and made it self-executing and removed the self-destruction, so it will use the `cwd` from where it's used, to be used by Dockerfiles that extend this base (think the `./Dockerfile` runs it)

stephenwf: The original `./patch-composer.php` is only used locally now, to get you a `./vendor` directory of the right modules for development

docker/madoc-omeka-s/usr/local/bin/madoc-installer
stephenwf: Any recommendations for the permissions, this script will be run as `www-data` so should own the directory at this point, no idea if Omeka uses the FS ever to list files or how it creates new files. Maybe `775` or `755` ?

garyttierney: `770` makes sense, I think. Users that aren't `www-data` have no business looking at these files.

PR #41, opened 2019-03-19
Performance improvements

stephenwf:
Summary of changes
  - All collections page now only loads 4 manifest per collection
  - Only 3 collections per page on all collections page
  - Viewing a single collection now only loads the manifests displayed
  - Viewing a single collection now only loads one canvas per manifest
  - Viewing a single manifest now only loads the canvases displayed

All requests to item relationships bypass Omeka to achieve this.


repos/iiif-storage/src/Controller/ManifestController.php
```diff
@@ -80,13 +81,14 @@ public function __construct(
...
+    {
         if ($collectionId) {
-            $collection = $this->collectionRepo->getById($collectionId);
-            if (!$this->collectionRepo->containsManifest($collection, $manifestId)) {
+            if (!$this->collectionRepo->containsManifest((int)$collectionId, $manifestId)) {
                 throw new NotFoundException();
             }
-            $collectionRepresentation = $this->collectionBuilder->buildResource($collection, $this->shouldUseOriginalIds());
+            $collection = $this->collectionRepo->getById($collectionId);
+            $collectionRepresentation = $this->collectionBuilder->buildResource($collection, $this->shouldUseOriginalIds(), 1, 1, 0);
```
stephenwf: 1 - manifest
1 - canvas
0 - offset

repos/iiif-storage/src/Extension/ImageSourceRenderer.php
```diff
@@ -87,13 +93,14 @@ public function renderCanvas(MediaRepresentation $image, ItemRepresentation $can
...
-            $manifests = $this->canvasRepository->getManifests($canvasRepresentation);
-
-            if (!empty($manifests)) {
+            if (!empty($manifestIds)) {
                 // @todo might be more than one manifest.
-                $manifestRepresentation = array_shift($manifests);
-                $manifest = $this->manifestBuilder->buildResource($manifestRepresentation);;
+                $manifestId = array_shift($manifestIds);
+                $manifestRepresentation = $this->manifestRepository->getById($manifestId);
+                $manifest = $this->manifestBuilder->buildResource($manifestRepresentation, false, 1, 1);
```
stephenwf: 1 - page
1 - perPage

repos/iiif-storage/src/Controller/ManifestController.php
garyttierney: I can't make sense of what's going on here. What's `buildResource()` do?

repos/iiif-storage/src/Utility/CheapOmekaRelationshipRequest.php
```diff
@@ -0,0 +1,141 @@
...
+    public function getEtag(int $resourceId)
+    {
+        $statement = $this->connection->prepare(self::ETAG);
+        $statement->bindValue('resourceId', $resourceId);
+        $statement->execute();
+
+        $result = $statement->fetch();
+
+        if (!$result) {
+            return null;
```
garyttierney: Wouldn't an exception be more appropriate?

repos/iiif-storage/src/Utility/CheapOmekaRelationshipRequest.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public function selectSource(int $fromId)
+    {
+        $dcTermsSource = $this->saturator->loadPropertyId('dcterms:source');
+        $sourceStatement = $this->connection->prepare(self::SELECT_SOURCE);
+        $sourceStatement->bindValue('termId', (int)$dcTermsSource, PDO::PARAM_INT);
+        $sourceStatement->bindValue('resourceId', (int)$fromId, PDO::PARAM_INT);
+        $sourceStatement->execute();
+        $sourceResult = $sourceStatement->fetch();
```
garyttierney: Can we verify that the result set only contains 1 record here?

repos/iiif-storage/src/Controller/ManifestController.php
stephenwf: So the various `buildResource` methods take an Omeka `ItemSetRepresentation` or `ItemRepresentation` and return a [wrapped representation](https://github.com/digirati-co-uk/madoc-platform/blob/feature/performance/repos/iiif-storage/src/Model/ManifestRepresentation.php) that gives access to:
- Original Omeka item and Omeka ID
- Underlying IIIF resource through IIIF-PHP
- Pagination if its a partial resource

It's a cleaner alternative to the `addMetadata` approach on IIIF-PHP objects used before to store things like the Omeka idea. This object is intended to be passed around and will probably grow as more data is built into manifests (such as statistics). 

Maybe the build resource should move to a factory. Something like:
```php
... ->buildResource(
    (new CollectionResource::fromItemSet($collection))
          ->limitManifestsLoaded(1)
          ->limitCanvasesLoaded(1)
          ->canvasOffset(0)
          ->useOriginalIds()
);
```

repos/iiif-storage/src/Utility/CheapOmekaRelationshipRequest.php
stephenwf: Yes good point, might be a good short-cut to a 404 in the controller too

repos/iiif-storage/src/Utility/CheapOmekaRelationshipRequest.php
stephenwf: Can't guarantee that, but its a computed field (the source JSON) so I've updated to close the cursor after getting the first result. It would take someone circumventing the CMS to add more than one JSON source though

garyttierney: What I mean is: is it fine if we just pick the first result in a scenario where 2 come back, or would we want to say "this shouldn't have been possible"

stephenwf: Yeh it's fine to pick just the first, if theres 2, someone has added that second field and should be ignored. Because the field is hidden from the UI by deafult, I don't think throwing an error here would be useful to an editor as they would not easily be able to fix the issue (or get in the situation in the first place)

PR #66, opened 2019-04-09
Added configurable pagination to collection and manifest pages

stephenwf:


repos/iiif-storage/Module.php
```diff
@@ -94,6 +94,38 @@ public function attachListeners(SharedEventManagerInterface $sharedEventManager)
                             ])
                         ->setValue($form->getSiteSettings()->get('original-ids', false))
                     )
+                    ->add(
+                        (new Text('collections-per-page'))
```
garyttierney: We should validate that the input coming in is a valid number (or you might end up casting the result to 0 and get 0 collections per page).

PR #68, opened 2019-04-09
Removed Transifex as dependency for localisation

stephenwf:
Split the configuration form into 2 options:
* Enable Localisation
* Enable Transifex

Also split the feature flags into 2, few updates to use the base config form for consistency thrown in too. Also commented out some dead code (according to Intelij)

### DIFF
--- a/repos/omeka-i18n-module/Module.php
+++ b/repos/omeka-i18n-module/Module.php
@@ -2,6 +2,7 @@
 
 namespace i18n;
 
+use Digirati\OmekaShared\ModuleExtensions\ConfigurationFormAutoloader;
 use i18n\Controller\LanguageSelectionController;
 use i18n\Event\TransifexProjectListener;
 use i18n\Event\TranslatableResourceListener;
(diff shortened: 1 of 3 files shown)


repos/omeka-i18n-module/Module.php
```diff
@@ -46,13 +55,17 @@ public static function isFeatureFlagEnabled(ContainerInterface $container)
         }
 
         $globalSettings = $container->get('Omeka\Settings');
-        $localizationSettings = $globalSettings->get('i18n', []);
+        $transifexEnabled = boolval($globalSettings->get('i18n_transifex-enabled', false));
```
garyttierney: Maybe forgetting something about Omeka setting namespacing, but what's with `_` then `-`?

stephenwf: It's a mistake on my part really, when I created:
https://github.com/digirati-co-uk/madoc-platform/blob/master/repos/shared-library/src/Framework/AbstractConfigurationForm.php#L46

I modelled it off of a few examples of configuration that already existed, which used `NAMESPACE_KEY_NAME` as the pattern, so I when automating the saving side, just followed that pattern. It might make more sense to be `NAMESPACE/KEY-NAME` as the convention though. It would require a SQL migration for existing sites with the configuration though.

PR #69, opened 2019-04-10
Locale configuration

### LINKED ISSUE #73: Ensure modules are translatable
Ensure string literals in custom templates and forms have `@translate` annotations or `translate()` functions around them. This will be tested with the extract tagged strings code that will extract from phtml, php and twig files.

stephenwf:
Closes #73 #80 #81 #82 #78


repos/shared-library/src/Utility/OmekaValue.php
```diff
@@ -20,6 +20,12 @@ public static function translateValue($representation, $term, $lang = '')
         if (!$representation instanceof ItemRepresentation && !$representation instanceof ItemSetRepresentation) {
             throw new LogicException('Only Items and ItemSets can be translated.');
         }
+
+        // Examples of input/output:
+        // en-gb -> en
+        // en-us -> en
+        // en    -> en
+        // cy-gb -> cy
         $primaryLang = explode('-', $lang)[0] ?? $lang;
```
garyttierney: We should use `Locale::parseLocale()` for this

stephenwf: Changed that now to use both the parse and comparisons

repos/iiif-storage/src/Aggregate/DereferencedManifest.php
```diff
@@ -70,11 +70,31 @@ public function mutate(ItemRequest $input)
...
+                    if ($summary) {
+                        $input->addFields(
+                            FieldValue::literalsFromRdf('dcterms:description', 'Summary', $summary)
+                        );
+                    }
+                }
+
+                $attribution = $input->getValue('sc:attributionLabel');
+                if (!$attribution || empty($attribution) || trim(current($attribution)->getValue()) === '') {
+                    $attribution = $manifest['attribution'] ?? null;
```
stephenwf: Need some abstraction here like:
```
$input->replaceIfEmpty( ... );
```
to avoid this verbose block being used all over the place

stephenwf:
Latest change:
Screenshot 2019-04-11 at 16 53 32

repos/iiif-storage/config/service.config.php
```diff
@@ -403,11 +428,15 @@
...
             },
+            'html' => function (ContainerInterface $c) {
+                return $c->get(HtmlIngester::class);
+            },
         ],
     ],
     'media_renderers' => [
         'invokables' => [
             'iiif' => null,
+            'html' => null,
```
garyttierney: Why is this needed? I see we're replacing it as a factory, does an invokable take precedence?

repos/iiif-storage/src/Model/BuiltCanvas.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace IIIFStorage\Model;
+
+
+class BuiltCanvas
```
garyttierney: Would be nice to call this `Canvas`, since it's built by `CanvasBuilder`, but that leaves us with a naming conflict across another namespace.

repos/omeka-i18n-module/src/Mvc/Router/SkippableSegment.php
```diff
@@ -0,0 +1,130 @@
...
+     */
+    public function __construct($route, array $constraints = array(), array $defaults = array(), array $skippable = array())
+    {
+        $this->defaults = $defaults;
+        $this->skippable = $skippable;
+        $this->parts = $this->parseRouteDefinition($route);
+        $this->regex = $this->buildRegex($this->parts, $constraints);
+    }
+    /**
+     * factory(): defined by RouteInterface interface.
```
garyttierney: Let's just do `@inheritdoc`

repos/omeka-i18n-module/src/Mvc/Router/SkippableSegment.php
```diff
@@ -0,0 +1,130 @@
...
+                case 'optional':
+                    $skippable = true;
+                    $optionalPart = $this->buildPath($part[1], $mergedParams, true, $hasChild, $options);
+                    if ($optionalPart !== '') {
+                        $path .= $optionalPart;
+                        $skip = false;
+                    }
+                    break;
+                case 'translated-literal':
+                    $path .= $translator->translate($part[1], $textDomain, $locale);
```
garyttierney: Are we using this? Could be potentially useful.

repos/omeka-i18n-module/src/Site/LocalizationListener.php
```diff
@@ -43,28 +45,44 @@ class LocalizationListener
...
 
         if (null === $locale || empty($locale)) {
             $locale = 'en';
         }
 
+        $isMultiLingual = $isOnSite ? boolval($this->siteSettings->get('i18n-multi-lingual-site')) : false;
+        /** @var TreeRouteStack $router */
+        $router = $event->getRouter();
+
+        if ($isMultiLingual && $routerLocale) {
```
garyttierney: Is setting it to null required if it's not present? If so we might as well capture this in a ternary: `$uriLocale = ($isMultilingual && $routerLocale) ? $locale : null;` 

Also, a nitpick: multilingual is one word, so shouldn't be uppercased at Lingual.

repos/shared-library/src/Utility/OmekaValue.php
```diff
@@ -0,0 +1,88 @@
...
+            if (self::langMatches($value->lang(), $lang)) {
+                $fallback = $value;
+            }
+        }
+
+        // Fallback found, then return that.
+        return $fallback;
+    }
+
+    static public function langMatches($langA, $langB)
```
garyttierney: Inconsistent order of modifiers here, `public static` and `static public` in this function.

repos/shared-library/src/Utility/OmekaValue.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        if ($langA === $langB) {
+            return true;
+        }
+        $valueLanguage = Locale::parseLocale($langA)['language'] ?? null;
+
+        // Check if they match each other, in either direction.
+        return (
+            $langA &&
```
garyttierney: Can we move this test for `$langA` up to an `if`  that exits early? Testing `$langA` isn't really part of the locale comparison and hurts the readability of this expression quite a bit.

repos/iiif-storage/src/JsonBuilder/MetadataAggregator.php
```diff
@@ -2,54 +2,63 @@
...
-                        break;
-                    case 'literal':
-                        $json['metadata'][] = [
-                            'label' => $value->property()->label(),
-                            'value' => $value->value(),
-                        ];
-                        break;
-                    default:
-                        break;
+                // @todo Pass the labels through the translator in Omeka
```
garyttierney: Hm, do we need to export vocabs to the external translation tool?

repos/iiif-storage/src/Model/BuiltCanvas.php
stephenwf: Once we drop the IIIF-PHP library, we could do that. Its strange because there's also `CanvasRepresentation` which wraps IIIF-PHP, JSON and the Omeka item which is passed to the templates!

repos/omeka-i18n-module/src/Mvc/Router/SkippableSegment.php
stephenwf: No currently, hadn't noticed the extra type

repos/iiif-storage/src/JsonBuilder/MetadataAggregator.php
stephenwf: Omeka 1.4 added [something that might help](https://github.com/omeka/omeka-s/pull/1373) but this appears to be just for displaying the vocabulary. It's possible we might need to do this, but I don't think it is worth doing at the moment. Specifically, the labels I'm referring to in this `@todo` are the ones from the resource templates:
Screenshot 2019-04-11 at 19 40 52

Which is used in the metadata key-value pair on an omeka item. (bad example here showing `sc:hasManifests` but imagine another RDF term)

So maybe we just need to export the resource templates, and then pass them through translate in the template

repos/iiif-storage/config/service.config.php
stephenwf: Possibly, I remember running into issues with the IIIF one, also to remove any ambiguity for anyone debugging. (Quick scan of Zend docs didn't produce an answer either)

repos/iiif-storage/src/JsonBuilder/MetadataAggregator.php
garyttierney: Ah yeah, that could be as simple as exporting message keys for `dcterms:identifier` or something.

PR #70, opened 2019-04-11
Added change log since 1.0.0

stephenwf:


garyttierney:
Any specific format we're following here? E.g., [Keep A Changelog](https://keepachangelog.com/en/1.0.0/).

Would be good to include it as a heading if so, like this: https://github.com/jenkinsci/analysis-model/blob/master/CHANGELOG.md#changelog

PR #84, opened 2019-04-12
Capture model translate

### LINKED ISSUE #72: Translate capture models inside of Omeka
At the moment you can translate capture models externally by providing a PO file for each. Since Omeka has added locale to the Omeka items, we should switch this over to use those and select the correct language values based on this when serving up the capture models.

The capture models will still use the header to request the correct locale to avoid canonical URL issues and generator values.

### LINKED ISSUE #74: Add locale to annotation studio configuration
At the moment there is not any configuration passed down to the annotation in the templates. We need to add `locale` option to the core component so that the context of the whole Annotation Studio changes to the configured language.

stephenwf:
## Added
- Capture models can now be translated inside of Omeka and will correctly be served depending on your locale (closes #72)

## Fixed
- Fixed bug where you could sync without Transfiex enabled
- Fixed bug where locale was not being passed to annotation studio (closes #74)

### DIFF
--- a/repos/annotation-studio/src/OmekaItemExpander.php
+++ b/repos/annotation-studio/src/OmekaItemExpander.php
@@ -3,8 +3,10 @@
 namespace AnnotationStudio;
 
 use Digirati\OmekaShared\Utility\OmekaValue;
+use Omeka\Api\Representation\AbstractRepresentation;
 use Omeka\Api\Representation\ItemSetRepresentation;
 use Omeka\Api\Representation\ValueRepresentation;
+use Omeka\I18n\Translator;
 use Zend\I18n\Translator\TranslatorInterface;
 
 class OmekaItemExpander
@@ -22,7 +24,7 @@ class OmekaItemExpander
     const FIELD_RENAMED_FROM = 'xx';
     const FIELD_RENAMED_TO = 'oa';
 
-    public static function expandChoices($items, callable $resolveUrl = null, TranslatorInterface $translator = null)
+    public static function expandChoices($items, callable $resolveUrl = null, Translator $translator = null)
     {
         $itemToReturn = [];
         /** @var $item ValueRepresentation */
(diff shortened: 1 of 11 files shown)


repos/annotation-studio/src/OmekaItemExpander.php
```diff
@@ -67,12 +69,21 @@ public static function expandDocument($document, callable $resolveUrl = null, Tr
...
-                        function ($data) use ($translator) {
-                            return self::toJsonValue($data, $translator);
-                        },
-                        $field
-                    );
+                    $locale = $translator->getDelegatedTranslator()->getLocale();
+                    $fallback = $first;
+
+                    foreach ($field as $singleField) {
+                        if ($singleField->lang() === $locale) {
```
garyttierney: `Locale::filterMatches(...)` might be a bit more "correct", though don't think it'll change the behaviour of this comparison.

repos/omeka-i18n-module/src/Controller/TranslationSyncController.php
```diff
@@ -40,12 +49,15 @@ public function showAction()
...
             [
                 'form' => $form,
+                'isEnabled' => $this->isEnabled,
             ]
         ))->setTemplate('admin/i18n-synchronize/show');
     }
 
     public function processAction()
     {
+        if (!$this->isEnabled) return null;
```
garyttierney: Hm, the typical approach with feature flagged web actions/controllers is to conditionally modify the routing tree so the controller doesn't need to be aware of the flag. That also helps identify places that are incorrectly linking to it (because generating a link to that route will now throw an exception).

repos/annotation-studio/src/OmekaItemExpander.php
garyttierney: Oh, I see you doing something maybe along those lines in langMatches.

stephenwf: Yeh, this will first check if its an exact match, then if not fallback to a fuzzy match (`en` vs `en_US`)

repos/omeka-i18n-module/src/Controller/TranslationSyncController.php
stephenwf: Been mulling over this one, I can't think of a straight nice way of doing this at the moment. The module is currently setting dynamic routes in `Module.php:151` but doesn't yet have access to the Omeka settings (from the database). Similar story with the navigation, it's all out of order.

I'm going to leave this as it is, but I do want to come back to this as a problem in general. In Omeka, the level of granularity of a feature flag is generally encapsulated in a module. So it may be the case that we split out the Transifex integration into its own module.

PR #100, opened 2019-06-10
Re-ingest image

stephenwf:
Example UI
Screenshot 2019-06-10 at 15 36 49

This is available at: `/admin/iiif/canvas/{id}` and the only action currently available is to request to re-request an image for thumbnails.

You can also POST directly to `/admin/iiif/canvas/{id}/ingest-thumbnail` if you wanted to batch some up.


stephenwf:
Change-log still needs to be updated. Will rebase 1.2 on-top of this for Stanford.

PR #104, opened 2019-06-11
Fixed MySQL max packet size error

stephenwf:


fractos:
worth adding `skip_name_resolve1 = 1   too

stephenwf:
I think in the `.ini` file it's just `skip-name-resolve` as the flag on the command line is `--skip-name-resolve `

>```opt_name```
>
>This is equivalent to --opt_name on the command line.
>
>```opt_name=value```
>
>This is equivalent to --opt_name=value on the command line. In an option file, you can have spaces around the = character, something that is not true on the command line. The value optionally can be 
 enclosed within single quotation marks or double quotation marks, which is useful if the value 
 contains a # comment character.

Source: https://dev.mysql.com/doc/refman/8.0/en/option-files.html

PR #124, opened 2019-10-31
Sorting room service

stephenwf:


config/nginx.conf
```diff
@@ -0,0 +1,60 @@
...
+            add_header Content-Type text/plain;
+            return 200;
+        }
+
+        try_files $uri @app;
+    }
+
+    error_log /var/log/nginx/error.log;
+    access_log /var/log/nginx/access.log;
+}
```
stephenwf: @garyttierney I was going to put this upstream into the Madoc docker image, but the locations are very specific to the modules in madoc. The base madoc image is really just madoc without any modules. Additions here compared to the base: added `presley` to the list of URLs that bypass the automatic `OPTIONS` check, since they support authorisation. There is also a static serving of the sorting room.

repos/iiif-storage/src/Aggregate/DereferencedManifest.php
```diff
@@ -45,7 +45,7 @@ public function mutate(ItemRequest $input)
     {
         foreach ($input->getValue('dcterms:identifier') as $field) {
             $manifestUrl = $field->getId();
-            if ($manifestUrl) {
+            if ($manifestUrl && $manifestUrl !== 'uri:internal') {
```
stephenwf: This isn't ideal, should possibly use a different identifier, or a completely different field to branch from. This is to bypass the validation from the resource template / class, since it requires a URL. It's valuable to have this validation, as it avoid manifests being imported incorrectly. However, it does make it a pain to make non-external manifests.

This could be fixed by splitting the resource templates into: `External IIIF Manifest` and `Custom IIIF Manifest` or something.

repos/iiif-storage/src/Controller/ManifestController.php
```diff
@@ -109,6 +109,8 @@ public function viewAction()
             $this->apiRouter->useOriginalUrls();
         }
 
+        $this->allowCors();
```
stephenwf: I believe this also fixed loading in the Universal Viewer

repos/iiif-storage/src/Controller/PresleyController.php
```diff
@@ -0,0 +1,168 @@
...
+                    $json['service'] = [
+                        "highlight" => "info",
+                        "profile" => "row-profile-here",
+                        "values" => [
+                            "",
+                            $json['label'],
+                            "",
+                            "",
+                            "",
+                            ""
```
stephenwf: This is a backwards compatible implementation. We'll start with this and then possibly move to using the metadata on the manifest to display in the Sorting Room.

repos/iiif-storage/src/Controller/PresleyController.php
```diff
@@ -0,0 +1,168 @@
...
+        unset($json['structures']);
+        unset($json['sequences']);
+        $json['@type'] = 'sc:Collection';
+        $json['@id'] = $this->url()->fromRoute(null, [], ['force_canonical' => true], true);
+        $json['members'] = array_map(function($manifest) {
+            return $this->builder->build($manifest, false, 0, 1)->getJson();
+        }, $manifests);
+
+        return new JsonResponse($json);
+    }
```
stephenwf: So the sorting room aims to split a big manifest, into smaller manifests. Sort of like:
```
- 300 page manifest
    - manifest 0-100
    - manifest 100-200
    - manifest 200-300
```

Since this doesn't really model in IIIF, the large manifest is _served_ as if it were a collection.

```
- 300 page manifest
    - "300 page manifest" collection [virtual]
        - manifest 0-100
        - manifest 100-200
        - manifest 200-300
```

The inner manifests all have a property `"dcterms:replaces": "https://example.org/manifest1"` pointing to the parent manifest. After creating the empty collection based on the manifest properties, it queries for manifests with `dcterms:replaces` that match itself. Put it all together and you have a collection.

You can do this for any manifest by going to:
```
/presley/collection/{manifest-id}
```

repos/iiif-storage/src/Controller/PresleyController.php
```diff
@@ -0,0 +1,168 @@
...
+            $item->addField(
+                FieldValue::entity('dcterms:replaces', $collectionId, 'resource:item')
+            );
+
+            foreach ($canvasIds as $canvasId) {
+                $item->addField(
+                    FieldValue::entity('sc:hasCanvases', $this->extractCanvasId($canvasId))
+                );
+            }
+        });
```
stephenwf: Nice and short implementation. This is not compatible with Presley, instead this is a way of creating a manifest from a label and list of canvas ids and then associating it with a "virtual collection" above.

```
POST /prelsey/collection/{manifest-id}/manifest
{
    "label": "label you want",
    "canvasIds": [
        "https://example.org/iiif/api/canvas/1",
        "https://example.org/iiif/api/canvas/2",
        "https://example.org/iiif/api/canvas/3",
    ]
}
```

repos/iiif-storage/src/Controller/PresleyController.php
```diff
@@ -0,0 +1,168 @@
...
+    public function extractCanvasId($id)
+    {
+        if (strpos($id, 'iiif/api') !== false) {
+            $parts = explode('/', array_shift(explode('?', $id)));
+            $id = array_pop($parts);
+            if ($id) {
+                return $id;
+            }
+        }
+        return null;
```
stephenwf: This is copied from another part of the code, it's an efficient way to convert Madoc-generated IIIF resources back to their Omeka id. It would be cleaner to use the Zend router to parse it out, but slower and effectively doing the same thing with regex.

garyttierney: Can we do this with a regex instead? This `pop(explode($str, shift(explode($str, $val)))` stuff is unreadable.

garyttierney: Also, for science: https://gist.github.com/garyttierney/f3267ca99ab10136445c0197ff505c61.

The regex is (unnoticeable) faster.

repos/iiif-storage/src/Controller/PresleyController.php
garyttierney: ... I don't want to know

docker-compose.yml
```diff
@@ -22,14 +23,33 @@ services:
...
       - OMEKA__SEARCH_INDEXER=search-indexer:8000
-    ports:
-      - "8888:80"
     volumes:
       - ./repos:/srv/omeka/repos
       - ./bare/omeka-s/application:/srv/omeka/application
       - ./files:/srv/omeka/files
       - ./config/omeka-app/application.config.php:/srv/omeka/application/config/application.config.php
       - ./translations:/srv/omeka/translations
+      - ./services/sorting-room/${SORTING_ROOM_DEV:-dist}:/srv/omeka/sorting-room
```
stephenwf: Needs to be removed

docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
       context: .
     links:
       - mysql
+      - sorting-room
```
stephenwf: to be removed

bin/madoc
```diff
@@ -141,6 +141,11 @@ case "${commandToRun}" in
         cp ./config/composer.json ./composer.json
         php ./config/patch-composer.php
         composer install --no-dev --no-scripts --optimize-autoloader
+
+        ## Sorting room
+        cd services/sorting-room;
```
garyttierney: This would be better as: `{cd services/sorting-room; ./build.sh}` since it runs in a subshell, and you don't have to `cd` back out when the subshell exits.

config/nginx.conf
```diff
@@ -0,0 +1,53 @@
...
+        try_files $uri @app;
+    }
+
+    location ~ ^/s/(.+)/auth/token {
+        try_files $uri @app;
+    }
+
+    location ~ /api/iiif-import {
+        try_files $uri @app;
+    }
```
garyttierney: These are unnecessary due to the fallback `location /` below.

repos/iiif-storage/src/Media/GenericServiceIngester.php
```diff
@@ -0,0 +1,58 @@
...
+            'raw-json' => json_encode($data, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_PRETTY_PRINT),
+        ];
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getLabel()
+    {
+        return 'Generic service';
```
stephenwf: Missing `@translate`

repos/iiif-storage/src/Utility/CheapOmekaRelationshipRequest.php
```diff
@@ -188,35 +197,90 @@ public function getManifestFromCanvas(int $canvasId)
...
     }
 
-    public function resourceExists(string $resource, string $id): bool
+    /**
+     * @param string $resource
+     * @param string $id
+     * @param string|null $baseUrl
+     * @return int | false
+     */
+    public function resourceExists(string $resource, string $id, string $baseUrl = null)
```
garyttierney: Should callers just be invoking this instead of `collectionExists`, etc? If not, might be better to make this non-public.

services/sorting-room/makeConfig.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+ENV_FILE_LOCATION="dist/assets/js/env.js"
+rm -f $ENV_FILE_LOCATION;
+
+/bin/cat > $ENV_FILE_LOCATION <<EOF
```
garyttierney: Do you get PATH errors without the `/bin/` prefix? Just `cat` would be better to handle systems that don't symlink /bin to /usr/bin.

services/sorting-room/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017
```
garyttierney: Missing copyright attribution (and wrong year)

services/sorting-room/makeConfig.sh
stephenwf: I don't recall, I think I may have had problems when using the alpine image.

garyttierney: `/usr/bin/cat` won't work on Alpine, but just `cat` will.

garyttierney:
```
> $ docker run -i -t alpine:latest sh                                                                                  
Unable to find image 'alpine:latest' locally                                                                           
latest: Pulling from library/alpine                                                                                    
c9b1b535fdd9: Pull complete                                                                                            
Digest: sha256:ab00606a42621fb68f2ed6ad3c88be54397f981a7b70a79db3d1172b11c4367d
Status: Downloaded newer image for alpine:latest                                                                       
/ # echo $PATH                                                                                                         
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin                    
/ # which cat                                                                                                          
/bin/cat     
```

PR #128, opened 2019-12-06
Translations fix

stephenwf:
- Fixes translations (incorrect `domain` value) which was working for the default site
- Adds `CHANGELOG.md` backlog


stephenwf:
`$madocTranslationPaths` is just a list of places to look for translations. It's primarily for navigation and other fixed content that cannot be easily translated directly in Omeka. The `default` just so happens to be the slug of the site added in by the vanilla Madoc install (and why this was difficult to replicate).

PR #131, opened 2020-01-28
User management

stephenwf:
### Added
- Added "display_errors" PHP configuration when `APP_ENV` is set to dev.
- Added new style for registration to match login in default theme
- Added ability to invite users to
- Added new UI for creating, and managing invitations
- Added ability to create non-public site and invite users to it, registering through the default site
- Added 4 new site-specific roles in preparation for upcoming changes
    - Limited Transcriber - Can transcribe tasks assigned to them.
    - Limited Reviewer - Can transcribe and review tasks assigned to them.
    - Transcriber - Can transcribe anything
    - Reviewer - Can transcribe and review anything
- Added ability to control how far down IIIF navigation each user role can go
- Added new fields to page blocks allowing you to target different site-roles with different blocks
- Added ability for existing users to be invited to new sites using the same links without re-registering
- Added automatic login when automatically activate is configured
- Added login or "view site" link to the thank you page, shown after registering

### Fixed
- Fixed bug in is registration permitted for non-public sites
- Fixed warning from undefined index when requesting user statistics
- Fixed various permissions and ACL surrounding non-public sites

-------------------------------------

In addition to these changes, there are a few further points:

- The ACL for Public User Module has been 100% rewritten
- This introduces a way to split `Module.php` into more manageable chunks.
- The ACLs for IIIF resources currently just take users to a page telling them they cannot see the resource. This will be changed shortly with additional options for users.
- Default Invitation settings are not controlled in the UI yet, and expiry is built in but does not currently show up in the UI. 
- There is some complexity when trying to invite users to a site that is not public. However the application will help to resolve this issue (by using a default site). 
- Admin UI will warn you when you are creating invites for a site with open registrations.

----------------------------------------

Screenshot 2020-01-28 at 16 41 01


repos/public-user/src/Module/GetConfig.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->config = Factory::fromFiles(
+            glob(__DIR__ . '/../../config/*.config.*')
+        );
+
+        return $this->config;
+    }
+
+    public function loadVendor()
+    {
+        if (file_exists(__DIR__ . '/build/vendor-dist/autoload.php')) {
```
garyttierney: Why do we still need to do this dance? We don't deploy modules without composer anymore.

repos/public-user/src/Module/SiteSettingExtension.php
```diff
@@ -0,0 +1,204 @@
...
+            ->setOptions([
+                'label' => $label,
+            ])
+            ->setValue($form->getSiteSettings()->get($id, $default));
+    }
+
+    public function extendSiteSettings(SiteSettingsForm $form)
+    {
+        $form->add(
+            (new Fieldset('public-user'))
```
garyttierney: Doesn't this produce an object that looks like:

```
[
    'public_user' => [
         'public-user-enable-registration' => ...,
    ]
]
```

Seems like you could omit the `public-user-` prefix in the fields themselves.

repos/public-user/src/Module/GetConfig.php
stephenwf: Will remove from all in separate PR (this was in due to splitting out of `Module.php`) and make sure it still works without

repos/public-user/src/Module/SiteSettingExtension.php
stephenwf: Unfortunately not, we could namespace them, like the configuration forms: https://github.com/digirati-co-uk/madoc-platform/blob/master/repos/iiif-storage/src/Admin/ConfigurationForm.php 

Same result in the end in terms of `name` props on the form, but much easier to scan and read.

garyttierney: Ah, these aren't configuration forms. Got it.

repos/public-user/src/Module/GetConfig.php
stephenwf: Ill try removing these and tidying them in separate PR, make sure that it will work correctly. I think we are using composer exclusively now.

repos/public-user/src/Module/SiteSettingExtension.php
stephenwf: Not the admin config forms, they are fieldsets in each of the individual sites. It would be nice if they could added to in a similar way with a similar API. Not sure exactly how yet though

repos/public-user/src/Module/SiteSettingExtension.php
```diff
@@ -0,0 +1,204 @@
...
+                        $form,
+                        'public-user-site-permissions-viewer-view-collection',
+                        'Can viewers see individual collection pages' // @translate
+                    ))
+                    ->add($this->createCheckbox(
+                        $form,
+                        'public-user-site-permissions-viewer-view-all-collections',
+                        'Can viewers see all collections page' // @translate
+                    ))
+                ->setAttribute('id', 'viewer-permissions')
```
stephenwf: @garyttierney do you think I should create a loop for these, instead of explicitly listing out the permissions. It's effectively a table of users and permissions. Each user and permission gets a checkbox. It's not the prettiest in the Omeka UI, but can possible come back to that. Was experimenting with `CheckboxList` too, which isn't styled in Omeka, but isn't too far off from being useful.

repos/public-user/src/Settings/PublicUserSettings.php
```diff
@@ -15,7 +14,22 @@ class PublicUserSettings
...
+        'admin' => 'Admin',
+        // Can transcribe tasks assigned to them.
+        'limited-transcriber' => 'Limited Transcriber',
+        // Can transcribe and review tasks assigned to them.
+        'limited-reviewer' => 'Limited Reviewer',
+        // Can transcribe anything
+        'transcriber' => 'Transcriber',
+        // Can transcribe and review anything.
+        'reviewer' => 'Reviewer',
+    ];
```
stephenwf: I thought about making this non-constant, but I think we want this to be a constant. We're not looking for infinitely expandable user roles at the moment, we want stable predictable roles that we can build customisable functionality around.

repos/public-user/src/Settings/PublicUserSettings.php
```diff
@@ -29,6 +43,32 @@ public function __construct(
         $this->currentSiteSettings = $currentSiteSettings;
     }
 
+    const APPLY_CUSTOM_SITE_PERMISSION_TO_ROLES = [
+        'viewer',
+        'limited-transcriber',
+        'limited-reviewer',
+        'transcriber',
```
stephenwf: In Omeka there are 3 levels of functionality, `admin`, `editor` and `viewer`. They are assigned numbers 3-1. Functionality and permissions are then checked using a simple greater/less than. Effectively `editor` gets all the viewer permissions and the editor ones, and admin gets them all. This is an extension of that, these are the permissions that need to be checked on top of this structure. Everything that's not an `admin` or `editor` has base level permissions of a `viewer`. This is one dimension of the permission model in the configuration, with the `CUSTOM_SITE_PERMISSIONS` being the other dimension.

repos/iiif-storage/src/Controller/CollectionController.php
```diff
@@ -86,6 +94,11 @@ public function viewAction()
 
         $this->paginateControls($viewModel, $collectionRepresentation->getTotalResults(), $this->getManifestsPerPage());
 
+        if (!$this->acl->userIsAllowed($this->getAclResource('iiif-collection'), 'view')) {
```
stephenwf: This is the current "shorthand" way of checking if a user has a permission. It's leaning on the `AbstractPsr7ActionController` for the `getAclResource` abstraction. Ideally I would like something like:

```php
$this->userIsAllowedOnSite('iiif-collection', 'view');
```
Or something, to make these calls easy to read and hard to get wrong.

repos/public-user/src/Module/AclRules.php
```diff
@@ -0,0 +1,181 @@
...
+
+        $acl->allow(
+            null,
+            'iiif-collection',
+            'view-all',
+            new UserHasSitePermission(
+                $settings,
+                'view-all-collections'
+            )
+        );
```
stephenwf: This needs some explaining. The role is null, since it applies to all roles, that one's easy. The `iiif-collection` is the resource that the permission is being applied to. It's incorrect to look at this in order to query the ACL however, since `iiif-collection` is a reference to the ID of the resource.

The resource is a wrapper around this ID. In order to have enough information to determine if the user has a site permission we need: the current site, the full user object and the resource being requested. The [wrapper](repos/shared-library/src/Framework/ResourceWrapper.php) accepts the user and site in its constructor.

It is constructed in  the [abstract controller](https://github.com/digirati-co-uk/madoc-platform/pull/131/files#diff-ed379415660c1c012ae213fb1832ca4cR29-R37) so can be accessed by any of the modules. 

When the `UserHasSitePermission` is checked, it takes the user and site from this wrapper and decides if the user can do the thing. It also has an optional resource that we're not using yet. We could pass in the IIIF resource itself and have extended ACLs based on that (For example, lift the rule if you created the IIIF resource under certain configurations). 

Since the wrapper is in the shared module and the IDs are just chosen constants, any module, including custom per-installation modules, can narrow the access to IIIF resources based on the site, user and IIIF module in a decentralised way. 

I think, if this works well, the `ResourceWrapper` could also be a model for other permissions, not specifically IIIF, down the line.

repos/public-user/src/Module/AclRules.php
```diff
@@ -0,0 +1,181 @@
...
+
+        // The default assertion for viewing resources.
+        $viewerAssertion = (new AssertionAggregate())
+            ->setMode(AssertionAggregate::MODE_AT_LEAST_ONE)
+            ->addAssertions([
+                new SiteIsPublicAssertion,
+                new HasSitePermissionAssertion('viewer'),
+            ]);
```
stephenwf: This taken from Omeka's own ACL and is the default assertion for wether a particular user can see a site. This was needed to support non-public sites.

repos/shared-library/src/Framework/RenderMedia.php
```diff
@@ -18,6 +21,16 @@ public function render(
...
 
+//        if (isset($data[AbstractIngester::SITE_ROLE_ID])) {
+//            /** @var User $user */
+//            $user = $media->getServiceLocator()->get('Omeka\AuthenticationService')->getIdentity();
+//            // Get the role from the site.
+//            $role = SitePermissionsHelper::userRoleForSite($user, $site);
+//            if (!in_array($role, $data[AbstractIngester::SITE_ROLE_ID])) {
+//                return '';
+//            }
+//        }
```
stephenwf: This is something I need to rethink. There's no currently obvious way to get the current site when rendering media. I did have a working implementing that grabbed the route match, but that seems like needless overhead. Probably missing something, or something already in the DI container has had the site injected in and I've missed it.

PR #139, opened 2020-04-01
Madoc TS

stephenwf:


services/gateway/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
...
+
+JWT_SOURCE_64=$(base64 < /openssl-certs/madoc.pub);
+
+IFS=',' read -ra ADDR <<< "$JWT_SERVICES"
+for i in "${ADDR[@]}"; do
+    echo $(cat "./apps/$i.json" | jq --arg value "$JWT_SOURCE_64" '.jwt_source = $value') > "./apps/$i.json"
+done
+
+
+./entrypoint.sh
```
stephenwf, line 13: @donaldgray @garyttierney This is a limitation of the way you configure tyk to verify tokens that it doesn't issue. You have put the public key into the configuration. To make this dynamic, I'm just setting this in the entrypoint on a set of configurations passed in. 

There's a small pile up of dynamic things in these configurations. I'm considering creating a lighter, simple JSON format for services to describe some of this. That way, for a specific configuration, like Ghent, the actual tyk configurations can be generated correctly (with proxy + host set up, along with this JWT). I'm picturing this generation being part of the config repo – not dynamic. So they can still be changed and customised as much as needed.

Does need some thought though, it's not the easiest to configure. Similar issue with expanding the shared postgres database, where N number of services want a database at runtime.

garyttierney, line 13: This looks like it should be an `exec`.

services/gateway/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -ex
+
+JWT_SOURCE_64=$(base64 < /openssl-certs/madoc.pub);
+
+IFS=',' read -ra ADDR <<< "$JWT_SERVICES"
+for i in "${ADDR[@]}"; do
+    echo $(cat "./apps/$i.json" | jq --arg value "$JWT_SOURCE_64" '.jwt_source = $value') > "./apps/$i.json"
```
garyttierney, line 9: An atomic move would be a better approach than doing the `cat` before the redirection kicks in (which immediately truncates the file).

services/madoc-ts/src/middleware/omeka-page.ts
```diff
@@ -0,0 +1,24 @@
...
+      headers: context.req.headers as any,
+    }).then(r => r.text());
+
+    const [header, footer] = response.split('<!--{{ content }}-->');
+
+    context.response.body = `
+      ${header}
+      ${context.omekaPage}
+      ${footer}
+      source: ${`${omekaUrl}/s/${context.params.slug}/_template`}
```
stephenwf, line 21: Should remove this.

services/gateway/entrypoint.sh
garyttierney: If you used an atomic move you could get rid of the `echo` and `cat`:

```
jq --arg {...} < "apps/$i.json" > "apps/$i.new.json"
if [ $? -eq 0 ]; mv "apps/$i.new.json" "apps/$i.json"; fi
```

stephenwf: Ah good shout! I was having real problems trying to get JQ to read and write to the same file. This might solve the formatting too (currently new lines are just... lost)

garyttierney: Yeah, if you redirect to a file with `>` the first thing bash does is truncates the file. So `cat $file > $file` produces an empty file.

services/certs/generate.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ -f "madoc.key" ]; then
+  echo "Private key already exists, skipping..."
+else
+  openssl genrsa -out madoc.key 2048
+fi;
+
+if [ -f "madoc.pub" ]; then
+  echo "Public key already exists, skipping..."
+else
+  openssl rsa -pubout -in madoc.key -out madoc.pub
```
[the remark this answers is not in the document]
stephenwf: ah good point, if only private key has been deleted?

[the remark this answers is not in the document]
donaldgray: Yeah, avoids situation where theres an 'old' public key that doesn't match the newly created private key.

services/gateway/entrypoint.sh
stephenwf: Fixed now 👍

PR #184, opened 2020-08-12
Frontend improvements from UX review

robmarch2:
Miscellaneous frontend fixes arising from UX review. See https://docs.google.com/document/d/1J9p4INqonzgjbY6owS4FjjvlwbMzhdAUFiwN2xJg9LI/edit# for full details plus my comments describing what I've done

### DIFF
--- a/services/madoc-ts/src/frontend/admin/pages/content/manifests/manifest-projects.tsx
+++ b/services/madoc-ts/src/frontend/admin/pages/content/manifests/manifest-projects.tsx
@@ -5,6 +5,7 @@ import { SmallButton } from '../../../../shared/atoms/Button';
 import { Link } from 'react-router-dom';
 import { useData } from '../../../../shared/hooks/use-data';
 import { createUniversalComponent } from '../../../../shared/utility/create-universal-component';
+import { SnippetLarge } from "../../../../shared/atoms/SnippetLarge";
 
 type ManifestProjectsType = {
   params: { id: string };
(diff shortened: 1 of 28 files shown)


services/madoc-ts/src/frontend/admin/pages/content/manifests/manifest-projects.tsx
```diff
@@ -40,14 +41,13 @@ export const ManifestProjects = createUniversalComponent<ManifestProjectsType>(
...
-            </Link>
-            <p>
-              <LocaleString>{project.summary}</LocaleString>
-            </p>
-          </div>
+          <SnippetLarge
+            label={<LocaleString>{project.label}</LocaleString>}
+            portrait={true}
+            subtitle={<LocaleString>{project.summary}</LocaleString>}
+            link={`./../../projects/${project.id}`}
```
robmarch2, line 48: I'm not happy with what I've had to do here. The previous `<Link to=...>` approach generated the `href` relative to the current site (i.e. `localhost:8888/s/default/madoc/admin`) whereas using the same expression here generated it relative to the root URL (i.e. `localhost:8888`). I really don't like the fact that I've had to give this component knowledge of how it's used in the routes

stephenwf: I think you should be able to use
```
link={`/projects/${project.id}`}
```

That will take you to the admin page for the projects. The absolute path are all under `/s/default/madoc/admin` for these components and then everything under `site` is `/s/default/madoc`. Unfortunately you can't route between them without a page-refresh as they are completely separate React apps.

Going to look at cross-linking more generally so you can jump from a project or collection page on the admin to the site and back again.

robmarch2: Unfortunately that doesn't work - the `a` tag ends up as:

```
<a role="link" href="/projects/1" class="sc-fzppAM fqNlnf">view project</a>
```

stephenwf: ahh, it's creating an anchor tag directly. I think if you add a prop: `linkAs={HrefLink}` (in the shared utilities) that will fix the issue 👍

PR #207, opened 2020-09-16
Added search API

stephenwf:


services/gateway/services/search.json
```diff
@@ -0,0 +1,28 @@
...
+    "versions": {
+      "Default": {
+        "name": "Default",
+        "expires": "3000-01-02 15:04"
+      }
+    }
+  },
+  "proxy": {
+    "listen_path": "/api/search",
+    "target_url": "http://search:8000/",
```
stephenwf, line 24: For now it will prefix the search, so `/indexables` in the service will be `/api/search/indexables`. If this changes in the service we can change the target url to have `api/search` on the end of it

ionab (approved):

PR #209, opened 2020-09-17
Update API to early exit if the search term is null

ionab:


services/madoc-ts/src/gateway/api.ts
```diff
@@ -394,6 +394,16 @@ export class ApiClient {
...
+        results: [],
+        pagination: {
+          page: 1,
+          totalResults: 0,
+          totalPages: 1,
+        },
+        facets: {},
+      };
     return await this.request<SearchResponse>(
       `/api/search/search?${stringify({ fulltext: searchTerm, page: pageQuery, facetType, facetValue })}`
```
stephenwf, line 408: Can we also chuck in a default page (1) I think that's causing an error on initial load

stephenwf (approved):

PR #217, opened 2020-10-06
MAD-373 - Update / delete options on manifests search index

ionab:


services/madoc-ts/src/frontend/admin/pages/content/manifests/edit-manifest-metadata.tsx
```diff
@@ -43,6 +43,21 @@ export const EditManifestMetadata: UniversalComponent<EditManifestMetadataType>
...
+        id: params.id,
+        thumbnail: manifest.manifest.thumbnail || '',
+        // how do I map here to the right shape for the resource
+        resource: { ...manifest.manifest, type: 'Manifest', id: params.id },
+        contexts: { contexts },
+      };
+
+      // // Reingest the search index
+      await api.searchReIngest(resource);
+
```
stephenwf, line 60: This will be called from the frontend - we could do the same step in the `update-metadata` on the backend - which would catch cases when the metadata it updated from the API

stephenwf (approved):

stephenwf (approved):

PR #223, opened 2020-10-21
Mad 359

ionab:
Awaiting confirmation on the A-Z, Z-A issue when it comes to international strings


stephenwf:
Looks good, but let's close for now until we resolve the requirement

PR #231, opened 2020-10-27
Add buttons in for the search facets

ionab:


services/madoc-ts/src/frontend/site/pages/search.tsx
```diff
@@ -1,127 +1,138 @@
...
+        newFacets[type].push(value);
+
+        history.push(
+          `${pathname}?${stringify({
+            fulltext,
+            page: 1,
+            madoc_id,
+            facets: JSON.stringify(newFacets),
+          })}`
+        );
```
stephenwf, line 118: @ionab I think the change I made will apply the facets as soon as you select the item, instead of storing in state first. Sorry! You could hold the state inside of the `SearchFacets` component and then pass down _all_ of the facets when you call `applyFacets`

ionab: Ok think I've got it working again now, storing the state locally.

stephenwf (approved):

PR #237, opened 2020-11-03
Search fixes

stephenwf:


ionab (approved):
Ah excellent, this was on my list to check with Matt this morning!

PR #242, opened 2020-11-05
Adds MetaData to ViewCanvas Page and ViewManifests Page

ionab:


services/madoc-ts/src/frontend/site/pages/view-canvas.tsx
```diff
@@ -22,9 +25,33 @@ import { CrowdsourcingTask } from '../../../gateway/tasks/crowdsourcing-task';
 import { CrowdsourcingReview } from '../../../gateway/tasks/crowdsourcing-review';
 import { SuccessMessage } from '../../shared/atoms/SuccessMessage';
 import { CanvasLoaderType } from './loaders/canvas-loader';
+import { TabPanel } from '../../shared/components/TabPanel';
+import { InternationalString } from '@hyperion-framework/types';
+
+import styled from 'styled-components';
+import { ManifestFull } from 'types/schemas/manifest-full';
```
stephenwf, line 32: I think this non-relative path might break a part of the build

services/madoc-ts/src/frontend/site/pages/view-canvas.tsx
```diff
@@ -223,6 +252,21 @@ export const ViewCanvas: React.FC<ViewCanvasProps> = ({
...
+      })();
+    }
+    if (collectionId) {
+      (async () => {
+        const { collection } = await api.getCollectionById(Number(collectionId));
+        setCollectionName(collection?.label);
+      })();
+    }
+  }, [manifestId, collectionId]);
+
```
stephenwf, line 269: Can you check the props in ViewCanvas - I believe the manifest and collection should be passed down.

ionab: We have `  project,
  manifestUserTasks,
  canvasTask,
  userTasks,
  canUserSubmit,
  canvas,
  isLoadingTasks,` and the manifest id and collection id from the params

stephenwf: Ah I don't think there added to the type
Screenshot 2020-11-06 at 15 33 50

ionab: ahh, ok so I should be able to access it and update the type?

services/madoc-ts/src/frontend/shared/components/MetaDataDisplay.tsx
```diff
@@ -0,0 +1,57 @@
...
+}> = ({ metadata, style }) => {
+  return (
+    <MetadataDisplayContainer style={style} >
+      {metadata.length < 1 ? 'No metadata to display' :
+      metadata.map((metadata: any) => {
+        return (
+          <MetadataContainer>
+            <MetaDataKey>
+              <LocaleString>{metadata.label}</LocaleString>
+              {' : '}
```
stephenwf, line 47: I'm not sure about the UX - but it's common for the metadata to be bold (and no ` : ` spacer). Display and layouts looking good 👍

ionab: Yeah I wasn't exactly sure on this either but might be a good candidate for UXPin iteration with Ville! Had intended on doing that today forgetting Ville doesn't work Fridays! I'll update to bold and remove the :

stephenwf (approved):

PR #328, opened 2021-04-29
MAD-779: Ghent UI changes

robmarch2:


services/madoc-ts/src/frontend/shared/atoms/Heading3.tsx
```diff
@@ -11,6 +11,13 @@ export const Heading3 = styled.h3<{ $margin?: boolean }>`
     css`
       margin-bottom: 1em;
     `}
+    
+  & > a {
+    transition: color 0.3s;
+    &:hover {
+      color: #3579f6;
```
robmarch2, line 18: I went for the same blue here as the primary buttons - not sure if this is theme-dependent though?

stephenwf:
Merged in #332

PR #335, opened 2021-05-06
MAD-794: Make header element visibility configurable

robmarch2:


stephenwf (approved):

stephenwf:
@robmarch2 sorry about that test - really need to change that one. Breaks every time we add config!

robmarch2:
> @robmarch2 sorry about that test - really need to change that one. Breaks every time we add config!

No worries, it's easy enough to find once you know it's there. Maybe a reminder to check the tests in the documentation would be enough on this one

PR #347, opened 2021-05-10
Added lower error boundary

stephenwf:


services/madoc-ts/src/frontend/shared/components/NotFoundPage.tsx
```diff
@@ -8,7 +9,7 @@ export const NotFoundPage: React.FC<{ error?: any }> = () => {
   return <h1>Not found</h1>;
 };
 
-export const ErrorPage: React.FC<{ error?: Error }> = props => {
+export const ErrorPage: React.FC<{ error?: Error; resetError?: () => void }> = props => {
```
robmarch2, line 12: This new prop looks unused to me. Am I missing something here?

PR #392, opened 2021-06-07
Move registration pages into typescript

robmarch2:


stephenwf:
Rolled this into #432

PR #395, opened 2021-06-11
MAD-749: Show preview and delete options for entity selectors

robmarch2:
Depends on https://github.com/digirati-co-uk/capture-models/pull/41

### DIFF
--- a/services/madoc-ts/src/frontend/shared/caputre-models/VerboseSelector.tsx
+++ b/services/madoc-ts/src/frontend/shared/caputre-models/VerboseSelector.tsx
@@ -1,7 +1,6 @@
-import { Revisions, RoundedCard } from '@capture-models/editor';
-import { useSelectorStatus } from '@capture-models/plugin-api';
+import { RoundedCard, EntityInstance } from '@capture-models/editor';
 import { BaseSelector } from '@capture-models/types';
-import React, { useCallback, useEffect } from 'react';
+import React from 'react';
 
 type VerboseSelectorProps = {
   selector: BaseSelector;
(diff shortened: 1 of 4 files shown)


services/madoc-ts/src/frontend/shared/caputre-models/VerboseSelector.tsx
```diff
@@ -21,7 +22,26 @@ export const VerboseSelector: React.FC<VerboseSelectorProps> = ({ selector, read
 
   useEffect(() => componentWillUnmount, [componentWillUnmount]);
 
+  const updateSelector = useCallback(
+    state => {
+      if (selector) {
+        selector.state = state;
+        updateSelectorValue({ selectorId: selector.id, state });
```
stephenwf, line 29: I think this may break in some cases, I think this would fit better in the `updateSelectorValue` method itself here:
https://github.com/digirati-co-uk/capture-models/blob/master/packages/editor/src/stores/revisions/revisions-store.tsx#L280-L334

So it can go through the existing tests.

robmarch2: The problem with this suggestion is that the selector that drives the component (i.e. `selector` here, which goes into `useSelectorStatus`) isn't the same object as the one being available in `updateSelectorValue` (which is looked up from the ID). When this change goes into capture models instead of here, the fix no longer works because the selector that drives the component has the wrong state on it

stephenwf:
@robmarch2 I merged https://github.com/digirati-co-uk/capture-models/pull/41 and tested, looks like it's working 👍 

I also copied it into madoc for now. The overhead of supporting those external components is a pain. I think most of `@capture-models/editor` will need a rethink, there's the `frontend/shared/capture-models` code in madoc-ts that might all get moved across.

stephenwf (approved):

PR #401, opened 2021-06-21
Update madoc.json

DavyVerbeke:
+358:  "Welcome back {{user}} !": "Welkom terug {{user}} !",

=> translation string not working?


stephenwf:
@DavyVerbeke the key for that one is `Welcome back {{user}}` without the exclamation point

PR #409, opened 2021-06-28
MAD-558 - Deleting projects, collections, manifests and canvases

stephenwf:
* Fact gathering for deletion
* Showing those statistics to users
* Placeholder for deletion steps in route

_**Can't be merged as it currently disables normal deletion.**_


services/madoc-ts/src/frontend/admin/pages/crowdsourcing/projects/delete-project.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <p>
+              This project contains <strong>{data?.collectionCount}</strong> collections, and
+              <strong>{data?.manifestCount}</strong> manifests. These will not be deleted.
+            </p>
+            {data?.siteCount !== 1 ? (
+              <p>
+                This project will still be available on <strong>{data.siteCount - 1}</strong> other site(s). You must
+                delete from all sites to fully delete the project.
+              </p>
+            ) : null}
```
stephenwf, line 40: Projects only belong to one site, so can be removed 👍

services/madoc-ts/src/routes/iiif/canvases/delete-canvas-summary.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const siteApi = api.asUser({siteId});
+
+  // Fact checking stage.
+  const { site_count, manifest_count } = await context.connection.one(
+    // This will let us know if the manifest appears on any other sites.
+    // If === 1 then we can safely delete underlying resource.
+    // TODO can a canvas appear in multiple manifests on the same site?
+    sql<{ site_count: number; manifest_count: number }>`
+      select COUNT(distinct site_id) as site_count, COUNT(*) as manifest_count from iiif_derived_resource where resource_id=${canvasId}
```
stephenwf, line 18: `canvas_count` in this file should probably be `canvas_count`

services/madoc-ts/src/routes/projects/delete-project-summary.ts
```diff
@@ -0,0 +1,67 @@
...
+    sql<{ site_count: number; manifest_count: number }>`
+      select COUNT(distinct site_id) as site_count from iiif_derived_resource where resource_id=${projectId}
+    `
+  );
+
+  const { collection_count } = await context.connection.one(
+    sql<{ collection_count: number }>`
+      select COUNT(*) as collection_count from iiif_derived_resource
+        where
+          id in (select item_id from iiif_derived_resource_items where resource_id = &{projectId})
```
stephenwf, line 25: The project contains a `collection_id` on its table. This is what the `resource_id` should match to find the manifests/collections. Easiest thing might be to fetch from the API: `const project = await siteApi.getProject(projectId)` and then `project.collection_id` is what you need

stephenwf: Also in template literals you need to use `${}` instead of `&{}`

services/madoc-ts/src/routes/iiif/canvases/delete-canvas-summary.ts
robmarch2: Do you mean `manifest_count` should be `canvas_count` instead? What I was trying to do here was work out how many manifests this canvas appears in

services/madoc-ts/src/routes/projects/delete-project-summary.ts
robmarch2: > The project contains a `collection_id` on its table. This is what the `resource_id` should match to find the manifests/collections. Easiest thing might be to fetch from the API: `const project = await siteApi.getProject(projectId)` and then `project.collection_id` is what you need

Ah yeah, good catch. In my local environment the project ID was always the same as it's collection ID, but that probably isn't a fair assumption in all cases.

> Also in template literals you need to use ${} instead of &{}

🤦 Oops.

services/madoc-ts/src/routes/iiif/canvases/delete-canvas-summary.ts
stephenwf: Ah right, yes that makes sense 👍

services/madoc-ts/src/database/queries/deletion-queries.ts
```diff
@@ -0,0 +1,14 @@
+import { sql } from 'slonik';
+
+export function deleteIiifMetadata(id: number) {
+  return sql<{ resource_id: number; total: number }>`
+    delete from iiif_metadata where resource_id = ${id};
+  `;
+}
```
stephenwf, line 7: There's 2 types of metadata. The original metadata (`resource_id = :id AND site_id is null`) and also site-specific overrides (`resource_id = :id and site_id = :sid`)

robmarch2: Yeah, I noticed that when doing the metadata deletion work. Locally this function is currently looking like this:

```
export function deleteIiifMetadata(id: number, siteId?: number) {
  if (siteId) {
    return sql`
      delete from iiif_metadata where resource_id = ${id} and id = ${siteId};
    `;
  }
  return sql`
    delete from iiif_metadata where resource_id = ${id};
  `;
}
```

stephenwf: Right that makes sense, so if you don't pass in a site ID you'll delete _all_ piece of metadata.

services/madoc-ts/src/utility/deletion-utils.ts
```diff
@@ -0,0 +1,8 @@
+import fs from 'fs';
+
+export function removeIiifFromDisk(path: string) {
+  if (!fs.existsSync(path)) {
```
stephenwf, line 4: I think we should add an additional check here to make sure the `path` starts with where we know the IIIF should be. Although unlikely, we don't want to delete things outside of that folder. 

The path can be found in a function: `saveManifestToDisk` in `services/madoc-ts/src/gateway/tasks/task-helpers.ts`

The path is:
```
${fileDirectory}/original/madoc-manifests/
```

stephenwf:
Looks good 👍  (can't approve, since I opened the PR)

PR #426, opened 2021-08-10
Add project export tab

robmarch2:


services/madoc-ts/src/routes/projects/export-project-template.ts
```diff
@@ -0,0 +1,34 @@
...
+    type: `template-${project.id}-${project.slug}`,
+    metadata: {
+      label: `Project ${project.id} export`,
+      description: `An export of project ${project.slug}`,
+      version: '1.0.0'
+    },
+    configuration: {
+      defaults: project.config,
+    },
+    setupModel: captureModel.document,
```
stephenwf, line 31: This can be removed, the set up model is a capture model that would be shown to the user when creating the project (so they can click some checkboxes to customise or something) and not yet used.

services/madoc-ts/src/routes/projects/export-project-template.ts
```diff
@@ -0,0 +1,34 @@
...
+  const { siteId } = userWithScope(context, []);
+  const { projectSlug, projectId } = parseProjectId(context.params.id);
+
+  if (!projectId && !projectSlug) {
+    throw new NotFound();
+  }
+
+  const userApi = api.asUser({ siteId });
+  const project = await userApi.getProject(context.params.id);
+  const captureModel: CaptureModel = await userApi.getCaptureModel(project.capture_model_id);
```
stephenwf, line 19: Regarding the potential ID conflicts. Here is the code for changing the IDs:
```js
import deepmerge from 'deepmerge';
import { generateId, traverseDocument } from '@capture-models/helpers';

// ... 


const newModel = deepmerge({}, model, { clone: true });
const updateId = (e: any) => {
  if (e.id) {
    e.id = generateId();
  }
};
traverseDocument(newModel, {
  visitEntity: updateId,
  visitField: updateId,
  visitSelector: updateId,
});
```

I _think_ this happens on the way in anyway, but maybe best just to be safe.

robmarch2: Would it be maybe worth instead clearing out the IDs here and regenerating them on the import?

stephenwf: Hmm, there's no type for ID-less capture models currently. I'd be wary of changing for one case. We can also change the IDs on the way in - if you wanted to use the same template. Having IDs may also let us use some of the existing components to preview what the model looks like before using it

robmarch2: Cool, well I've updated them here for now. There's a non-zero (if vanishingly small) chance that the regenerated IDs could clash when exporting a project from one madoc instance to another - which I assume is the primary use case here - so for safety's sake it may be worth regenerating them on the way in too. In the very unlikely event that the regeneration here picks out a value that already exists in the target environment you could simply reload the export page and try a second time though

stephenwf (approved):

PR #427, opened 2021-08-12
Remove unused code

robmarch2:
Unused code identified with [ts-prune](https://github.com/nadeesha/ts-prune)


stephenwf:
Looks good, nothing in there that looks like it is actually used. There are a few bits (like the activity stream types) that describe a full specification but we are not using it yet. I think we should leave those in. I'll leave comments on those.

services/madoc-ts/src/frontend/site/features/ReviewerSubmissionOverview.tsx
```diff
@@ -1,65 +0,0 @@
...
-                  </SmallButton>
-                ) : null}
-              </TableActions>
-            </TableRow>
-          ))}
-        </TableContainer>
-      ) : null}
-    </>
-  );
-};
```
stephenwf, line 65: I think this is fine to delete, but does raise a gap. Currently we don't show:
```
task.state. changesRequested
```
On the canvas page. Might need a JIRA ticket for that one.

services/madoc-ts/src/frontend/themes/helpers/themeVariable.ts
```diff
@@ -18,32 +18,3 @@ export function themeVariable<
...
-> {
-  const map: any = {};
-  const keys = Object.keys(definition);
-
-  for (const key of keys) {
-    map[key] = themeVariable(name, key as any, (definition as any)[key]);
-  }
-
-  return map;
-}
```
stephenwf, line 49: This is a short hand that's not been used yet. Currently you can create a theme variables for stylesheets.

Theme variables lets you define a whole bunch of variables at once:
```js
const test = themeVariables('accent', {
  primary: {
    default: 'white'
  },
  secondary: {
    default: 'gray'
  }
});

// test.primary
// test.secondary
```

Which is the same as the current helper, but with less code:
```js
const primary = themeVariable('accent', 'primary', {
  default: 'white',
});
const secondary = themeVariable('accent', 'secondary', {
  default: 'gray',
});
```

services/madoc-ts/src/utility/linking-property-types.ts
```diff
@@ -30,11 +30,3 @@ export function isLinkCaptureModelParagraphs(link: ResourceLinkResponse) {
...
   );
 }
-
-export function isLinkLocalOcr(link: ResourceLinkResponse) {
-  return isLinkPlaintext(link) || isLinkCaptureModelParagraphs(link);
-}
-
-export function isLinkOcr(link: ResourceLinkResponse) {
-  return isLinkPlaintext(link) || isLinkHocr(link) || isLinkMetsAlto(link);
-}
```
stephenwf, line 40: I'm surprised we don't use these! Fine to delete though

services/madoc-ts/src/frontend/site/features/ReviewerSubmissionOverview.tsx
robmarch2: Cool, yeah I agree this should probably be captured in a ticket. I've left it deleted for now

services/madoc-ts/src/utility/linking-property-types.ts
robmarch2: So I used ts-prune to find unused code, but there were quite a lot of false positives in the list it gave me. I ended up going through the list and checking in Intellij (which is better at detecting if code is actually used or not, but doesn't have a view to show all instances of unused code), and these were still unused. Happy to put these back in if you think they might be useful in the future though

stephenwf (approved):

services/madoc-ts/src/utility/linking-property-types.ts
stephenwf: No, should be fine to remove. I think we use the more granular ones that this uses

PR #508, opened 2022-01-17
Fixed searching by type

stephenwf:
performance regression

TODO: Changelog


mattmcgrattan (changes requested):
The query is incorrect. I've tested again locally, with a large SBB dataset. 

The correct query should be:

`type__iexact: searchQuery.iiif_type`

Just plain "type" rather than "iiif__type" or "indexables__iiif__type".

PR #527, opened 2022-03-08
Extract metadata editor as hook

stephenwf:


services/madoc-ts/src/frontend/shared/hooks/use-metadata-editor.ts
```diff
@@ -0,0 +1,289 @@
...
+      state.hasChanged = true;
+      state.fields[action.payload.id].value = action.payload.value;
+      if (state.modified.indexOf(action.payload.id) === -1) {
+        state.modified.push(action.payload.id);
+      }
+      break;
+    }
+
+    case 'CREATE_ITEM': {
+      console.log(action);
```
ionab, line 81:
```suggestion
```

ionab (approved):

services/madoc-ts/src/frontend/shared/hooks/use-metadata-editor.ts
stephenwf: Good catch, thanks!

PR #585, opened 2022-08-08
set init page number

Heather0K:
Not 100% sure this is the correct solution but, I think there is no initial page number being set and its coming back null or undefined so not attempting to load any images.


services/madoc-ts/src/extensions/capture-models/MediaExplorer/MediaExplorer.tsx
```diff
@@ -27,7 +27,7 @@ export const MediaExplorer: React.FC<MediaExplorerProps & {
   const { data: pages, fetchMore, canFetchMore, isFetchingMore } = useInfiniteQuery(
     ['media-explorer', {}],
     async (key, _, vars: { page?: number }) => {
-      return api.media.listMedia(vars.page);
+      return api.media.listMedia(1);
```
stephenwf, line 30: The page is set to `1` if `undefined` is [passed](https://github.com/digirati-co-uk/madoc-platform/blob/feature/fix-storybook-v2/services/madoc-ts/src/extensions/media/extension.ts#L82) and also defaulted on the [backend](https://github.com/digirati-co-uk/madoc-platform/blob/feature/fix-storybook-v2/services/madoc-ts/src/routes/media/list-media.ts#L7). The `vars.page` should be the current page from the `getFetchMode` below. It's possible it's not making the initial call because of something wrong there?

Heather0K: Yeah, It's only running that `getFetchMore` if I pass in a number in ` return api.media.listMedia(1);` Maybe its something to do with how the promise is constructed?

PR #608, opened 2022-11-09
Feature/contributions block

Heather0K:
Screenshot 2022-11-09 at 16 24 25


services/madoc-ts/src/frontend/site/features/ProjectContributors.tsx
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  const map = data?.tasks.map(task => task.assignee?.name);
+  const contributors = [...new Set(map)];
+
+  if (!data) return null;
+
+  return (
+    <ContributersWrapper>
+      <p>contributions by</p>
```
stephenwf, line 50: This might be clearer as a heading, maybe an h3? It's also something we might want a `t()` around

stephenwf:
@Heather0K this PR seems to delete the default Madoc theme. The theme directory is ignored, so you may need to run:
```
git add -f ./services/madoc-ts/themes/default-madoc-theme
```
to get the files back.

stephenwf (approved):

PR #609, opened 2022-11-10
MAD-1176 Button theme

Heather0K:


stephenwf:
I think we need more configuration options for the buttons. There are functional buttons, that should probably not be affected (I can see you've addressed them) 

* Success button
* Error button
* Error primary button

And then we have 2 buttons that could be independently styled with the theme:

* `<Button />`
* `<Button $primary/>`

Since we use them inline we might need some base button properties:
* `buttons.padding`
* `buttons.borderWidth`
* `buttons.borderStyle`
* `buttons.borderRadius`
* `buttons.boxShadow`

And then perhaps some for the button and primary button for active, hover and focus:
* `background`
* `color`
* `boxShadow`
* `borderColor`

Which, for background, could look like:
* `defaultButton.background`
* `defaultButton.hoverBackground`
* `defaultButton.activeBackground`
* `defaultButton.focusBackground`
* `primaryButton.background`
* `primaryButton.hoverBackground`
* `primaryButton.activeBackground`
* `primaryButton.focusBackground`

So it would be quite an involved customisation. It's also possible, due to performance of so many variables, that we use CSS variables - although that could be done later and more generally. 

The accessible color helper is good for editorial changes, but it's quite a costly check to happen on all renders unfortunately - and we can just ask for the colors in the theme and avoid calculating anything at runtime

services/madoc-ts/src/frontend/shared/navigation/Button.tsx
```diff
@@ -203,37 +249,42 @@ export const Button = styled.button<{
...
-  &:hover {
-    // background:{buttonColor};
-    // border-color: buttonColor};
-    // color:{BtnColor};
+  &:active {
+    box-shadow: inset 0 0 4px 0 ${defaultColor};
   }
 
-  &:focus {
+  &:focus &:focus-visible {
```
stephenwf, line 267:
```suggestion
  &:focus, 
  &:focus-visible {
```

PR #625, opened 2022-12-12
WIP tried to add event lisener onto selector from ViewField

Heather0K:


Heather0K:
First part of this done - selectors are highlighting when hovering over item in 'my contributions' 

But the other side where hovering over a selector should highlight item in 'my contributons' ; 

`helper.withSelector(selectorId).addEventListener('onPointerEnter', () => handleHover());`
is not working as expected. might be another bug?

stephenwf:
@Heather0K yes I think it might be a missing piece that's not passed down correctly to the selector. I think the first part is the main bit - maybe we can just go with that and create a ticket for the backlog for the hover in the backlog

stephenwf (approved):

PR #635, opened 2023-01-12
Begin matching types to md doc

Heather0K:


stephenwf (approved):
Looks good, seems like we are close to removing all the `compat*()` functions

PR #639, opened 2023-01-26
change validModels to filter non approved and empty fields

Heather0K:
Match logic in `/document-panel.tsx` for showing notification as `Viewdocument.tsx` uses to show approved submisions


services/madoc-ts/translations/en/madoc.json
```diff
@@ -473,6 +473,7 @@
   "No contributions in review": "No contributions in review",
   "No contributions yet": "No contributions yet",
   "No document yet": "No document yet",
+  "No dovcument yet": "No dovcument yet",
```
stephenwf, line 476:
```suggestion
```

stephenwf (approved):

PR #646, opened 2023-02-02
show highlight WIP

Heather0K:
Screenshot 2023-02-08 at 17 41 09

Made it purple so its different from the rest


services/madoc-ts/src/frontend/shared/hooks/use-read-only-annotations.ts
```diff
@@ -96,10 +104,32 @@ export function useReadOnlyAnnotations(isModelPage = false): ReadOnlyAnnotation[
         }
       }
     }
-
     return { regions, ids };
   }, [data, showDocumentRegions, styles.contributedDocument, annotations.ids]);
 
+  const unstyledRevisions = useMemo(() => {
+    const regions: ReadOnlyAnnotation[] = [];
+    const ids: string[] = [...annotations.ids];
```
stephenwf, line 112: Instead of `annotation.ids` we could store a list of revision ids. That should be available from `captureModel.revisions`. That would allow us to filter revisions by the creator. (`useUser()` to get user numeric ID and would compare on a urn like: `urn:madoc:user:123`).

The visitSelector() could then check `selector.revision` and only add to the list if it matches one of those revision ids.

stephenwf (approved):

PR #650, opened 2023-02-07
Mad 1244/embed block

Heather0K:
possible solution for [MAD-1244](https://digirati.atlassian.net/browse/MAD-1244)

Screenshot 2023-02-07 at 16 29 48
Screenshot 2023-02-07 at 16 29 14

[MAD-1244]: https://digirati.atlassian.net/browse/MAD-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


services/madoc-ts/src/frontend/site/features/EmbedItem.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }
+`;
+export const EmbedItem: React.FC<{
+  link?: string;
+  height?: string;
+  width?: string;
+}> = ({ link, height, width }) => {
+  return (
+    <EmbedWrapper>
+      <iframe src={link} height={height} width={width} />
```
stephenwf, line 17: For security we should add some extra properties to limit what the iframe has access to.
- [`sandbox=""`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox)
- [`referrerpolicy="no-referrer"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-referrerpolicy)

There may be others we could add, but I think limiting what the frame can do would be enough for now. We could add options to the block later too.

Heather0K: `sandbox="" `dosent allow any scripts so most embeded widgets wont do much I added` sandbox="allow-scripts" `instead?

Heather0K:
Screenshot 2023-02-07 at 17 54 35

services/madoc-ts/src/frontend/site/features/EmbedItem.tsx
Heather0K: I can remove this block if theres no usecase for it?

stephenwf (approved):

services/madoc-ts/src/frontend/site/features/EmbedItem.tsx
stephenwf: I think it still might be useful!

PR #653, opened 2023-02-09
Hotfix/reviewer dash bugs

Heather0K:
- Fix server errror when navigating 
- Add styled empty state for no review 
- Change langage on edit and focus buttons 
- Add `keepPreviousData` to `useInfiniteData` to smooth loading on sort 
- Wrap `single-reveview` in `MaximiseWindow` to keep `isOpen` state even when loading 
- add `refetch()` to `BeforeNavigate` to load next page


services/madoc-ts/src/frontend/shared/hooks/use-data.ts
```diff
@@ -186,7 +186,7 @@ export function useInfiniteData<Data = any, TKey = any, TVariables = any>(
       }
       return undefined as any;
     },
-    { ...(config || {}), useErrorBoundary: true }
+    { ...(config || {}), useErrorBoundary: true, keepPreviousData: true }
```
stephenwf, line 189: I think this might cause stale data issues in other places, you should be able to pass it into individual hooks if required.

services/madoc-ts/src/frontend/site/pages/tasks/review-listing/ReviewNagivation.tsx
```diff
@@ -131,7 +137,7 @@ export const ReviewNavigation: React.FC<{
             if (handleNavigation) {
               e.preventDefault();
               if (pages[pg].tasks) {
-                handleNavigation(pages[pg].tasks[idx + 1].id, pg + 1);
+                handleNavigation(pages[pg].tasks[idx + 1].id, pg + 1, idx + 2 > pages[pg].tasks.length - 1);
```
stephenwf, line 140: `tasks[idx + 1]` might be undefined if it's the last time in a page

services/madoc-ts/src/frontend/site/pages/tasks/review-listing/review-listing-page.tsx
```diff
@@ -312,6 +314,7 @@ function SingleReviewTableRow({ task, active, page }: { task: CrowdsourcingTask;
 
   return (
     <ThickTableRow
+      tabIndex={0}
```
stephenwf, line 317: There's a [hook in Manifest Editor](https://github.com/digirati-co-uk/iiif-manifest-editor/blob/583b28cd14aedadf08776e5c677b9a344f4b95ab/manifest-editor/src/hooks/use-keyboard-list-navigation.ts) that we could use here. 

It works like this:
```jsx
const containerProps = useKeyboardNavigation('data-something-unique');

return (
  <div {...containerProps}>
    <div tabIndex={0} data-something-unique="0">First</div>
    <div tabIndex={0} data-something-unique="1">Second</div>
    <div tabIndex={0} data-something-unique="2">Third</div>
  </div>
);
```

and will allow you to use the arrow keys to navigate the list.

services/madoc-ts/src/frontend/site/pages/tasks/review-listing/ReviewNagivation.tsx
Heather0K: It checks first to see if its getting the next item in the current page

` idx < pages[pg].tasks.length - 1 ? `

then it would go to `getNextPage()` where it gets the first item in the next page

stephenwf (approved):

PR #675, opened 2023-03-06
Feature/topic curation

Heather0K:
-


services/madoc-ts/src/extensions/enrichment/models.ts
```diff
@@ -12,10 +12,14 @@ export const entityTypeModel: CaptureModelShorthand<EnrichmentEntityType> = {
 
 export const entityModel: CaptureModelShorthand<EnrichmentEntity> = {
   other_labels: { type: 'international-field', label: 'Display label' },
-  label: { type: 'text-field', label: 'Canonical label' },
+  label: { type: 'text-field', label: 'Canonical label (this will effect the slug)' },
```
stephenwf, line 15: There is a "description" property that might be good for the extra information here (and in the other model)

stephenwf (approved):

PR #678, opened 2023-03-09
allow extra nav items to be added to nav

Heather0K:


services/madoc-ts/src/frontend/site/features/GlobalMenuStack.tsx
```diff
@@ -89,6 +93,12 @@ blockEditorFor(GlobalMenuStack, {
       fullWidth: false,
     },
     maxWidth: null,
+    slug1: '',
+    text1: '',
+    slug2: '',
+    text2: '',
+    slug3: '',
+    text3: '',
```
stephenwf, line 101: @Heather0K this might be a good place to test a simple case of the `allowMultiple` and I can jump in to fix the capture model short hand. It should allow users to make as many links as they need.

stephenwf (approved):
Language files look like they've got some extra fields, otherwise looks good 👍

PR #681, opened 2023-03-21
Updates to the indexing of manifests/canvases, and DBPedia docker-compose

mattmcgrattan:


docker-compose-dbpedia.yml
```diff
@@ -0,0 +1,255 @@
+version: "3.2"
+
+services:
+  # Internal services
+
+  gateway:
+    build:
+      context: services/gateway
+      dockerfile: Dockerfile.dev
```
stephenwf, line 9: I think the compose file could use the [extends](https://docs.docker.com/compose/extends/) pattern and just have the `spotlight.en` service. From a quick glance it looks like everything else is unchanged?

stephenwf (approved):

PR #699, opened 2023-04-04
Manifest enrichment

stephenwf:


services/madoc-ts/src/routes/enrichment/manifest-enrichment-pipeline.ts
```diff
@@ -0,0 +1,54 @@
...
+        const enrichmentPlaintext = await siteApi.enrichment.getEnrichmentPlaintext(first);
+        invariant(enrichmentPlaintext, 'Missing plaintext from enrichment');
+        if (enrichmentPlaintext.plaintext) {
+          const canvasId = parsed.id; // ??
+          return await siteApi.updateCanvasPlaintext(canvasId, enrichmentPlaintext.plaintext);
+        }
+      }
+    }
+  },
+};
```
stephenwf, line 54: This is the main part (other parts mostly filling gaps in Madocs API).

- `manifestEnrichmentPipeline` is the API route handler when an admin hits "Enrich"
  - Generates a webhook URL (12-hour token)
  - Creates the enrichment task
  - Returns task (@mattmcgrattan it would be useful to omit `callback_url` from the state in the future)
- `manifestEnrichmentPipelineEvent` is a short description of the webhook "type" and the fields expected in the response.
- `manifestEnrichmentHook` this is the function that is called when the task is complete. We get the webhook post-body JSON and an instance of the siteApi (already mapped correctly to the right site).
  - Fetch the enrichment task
  - Validate that it has a subject + is complete
  - If it's an `ocr_madoc_resource`:
    - Parse + validate the subject
    - check for `ocr_resources` (@mattmcgrattan will need to change if there are more than one here)
    - Fetch the plaintext + validate what we expect to see
    - Attach the plaintext to the canvas.

services/madoc-ts/src/extensions/enrichment/extension.ts
```diff
@@ -79,7 +79,28 @@ export class EnrichmentExtension extends BaseDjangoExtension {
...
+    return this.api.request<EnrichmentTask>(`/api/enrichment/tasks/madoc_manifest_enrichment_pipeline`, {
+      method: 'POST',
+      body: {
+        task: {
+          subject: `urn:madoc:manifest:${id}`,
+          parameters: [{ callback_url: callback }],
+        },
+      },
+    });
   }
```
stephenwf, line 104: This is the call to kick off the enrichment pipeline

Heather0K:
you might want to put a `word-break: break-all;` [here](https://github.com/digirati-co-uk/madoc-platform/blob/feature/IDA-1055-enrich-manifest/services/madoc-ts/src/frontend/admin/pages/system/webhooks/list-webhooks.tsx#L64) 

Screenshot 2023-04-04 at 18 13 27

PR #704, opened 2023-04-06
Footer-Grid improvments and nesed block form 

Heather0K:


stephenwf:
@Heather0K it looks like the page block data might have changed. Would it be possible to also support the old-props? Should be able to use `mapToProps` and `mapFromProps` to ensure the existing page blocks don't break

PR #714, opened 2023-04-20
Ns 17/hotfix/review issues

Heather0K:


services/madoc-ts/src/routes/projects/update-revision-task.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  if (!id) {
+    throw new Error('Task could not be updated');
+  }
+
+  if (!taskBody) {
+    throw new Error('Task could not be updated, no body');
+  }
+  await userApi.updateTask<CrowdsourcingTask>(id, taskBody);
+  context.response.status = 200;
```
stephenwf, line 19: This is fine, but we need to do some more validation to ensure:

- The user is allowed to update the specified task (reviewer/limited reviewer)
- The task is definitely a crowdsourcing task (fetch first)

stephenwf: There is also a `RequestError` that can be thrown that will make the status code 400 (indicating user error).

services/madoc-ts/src/routes/projects/update-revision-task.ts
```diff
@@ -0,0 +1,36 @@
...
+  const { siteId } = userWithScope(context, ['models.contribute', 'models.revision']);
+
+  const canCreate = userCan('models.create', context.state);
+  const canReview = userCan('models.revision', context.state);
+
+  const userApi = api.asUser({ siteId });
+  const id = context.params.taskId;
+  const taskBody = context.requestBody.task;
+
+  const task = await userApi.getTask(id);
```
stephenwf, line 17: One last thing, the user _might_ be able to edit other users contributions. I think the task has an `assignee` property that we could check if they `canCreate` but can't review. 

There is also `delegated_owners` which _should_ contain the assigned reviewer. (But I can't remember if that's 100% working, it should allow reviewers who are assigned to the review task to progress the "delegated" task, which is crowdsourcing task)

Heather0K: I think thats checked in [update-revision.ts ](https://github.com/digirati-co-uk/madoc-platform/blob/NS-17/hotfix/review-issues/services/madoc-ts/src/capture-model-server/routes/update-revision.ts) ?

services/madoc-ts/src/routes/projects/update-revision-task.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (limitedReviewer) {
+    if (task.delegated_task) {
+      const reviewTask = await userApi.getTask(task?.delegated_task, { all: true });
+      const assignedReviewer = reviewTask.assignee ? extractIdFromUrn(reviewTask.assignee.id) : '';
+      if (assignedReviewer !== currentUser) {
+        throw new Error('Not authorised');
+      }
+      await userApi.updateTask<CrowdsourcingTask>(id, taskBody);
+      context.response.status = 200;
```
stephenwf, line 44: Need a `return;`  statement here or it will immediately fall through to the throw error

stephenwf (approved):

PR #715, opened 2023-04-20
check if max contirbutions reached and prevent further sumision

Heather0K:


stephenwf:
I think this looks fine. The only case I can think is if the current user has a submission - then it shouldn't be marked like that, but I think that should be covered in the UI elsewhere

PR #726, opened 2023-05-10
Mad 1123/feature/no item page

Heather0K:


services/madoc-ts/src/frontend/site/pages/loaders/canvas-loader.tsx
```diff
@@ -29,6 +30,15 @@ export const CanvasLoader: UniversalComponent<CanvasLoaderType> = createUniversa
       [data]
     );
 
+    if (!data || data.canvas.source_id === 'not-found') {
```
stephenwf, line 33: When the data is loading, `data` will be undefined, so as you are navigating it may show "not found" and then jump to a found page. 

the `userData` does return a status. I think this should only be used _if_ we know it's an error (so data source id matching, or status is 'error')

If we don't know, and it's still loading, we can optimistically render what we had before.

The reason is that `<AutoSlotLoader />` will kick off requests to get the slots, and `<Outlet />`  will also allow other loaded to run in parallel. For example, if you're viewing a canvas within a project, the loaders will look something like this:

```jsx
<ProjectLoader>
  <ManifestLoader>
    <CanvasLoader>
       ...
    </CanvasLoader>
  </ManifestLoader>
</ProjectLoader>
```

When that page is first loaded, because all the components optimistically render the `<Outlet/>` they all fire their requests in parallel for the project, manifest and canvas.

If we don't render the `<Outlet/>`, then it will waterfall. So first it will wait for the project to be loaded (as `data === undefined`):
```jsx
<ProjectLoader>
  ...
</ProjectLoader>
```
And then the manifest:
```jsx
<ProjectLoader>
  <ManifestLoader>
    ...
  </ManifestLoader>
</ProjectLoader>
```
And so on.

services/madoc-ts/src/routes/iiif/manifests/get-manifest.ts
```diff
@@ -46,7 +46,8 @@ export const getManifest: RouteMiddleware<{ id: string }> = async context => {
 
   const manifest = table.manifests[`${manifestId}`] || {
     id: manifestId,
-    label: { none: ['Untitled manifest'] },
+    label: { none: ['Manifest not found'] },
+    source: 'not-found',
```
stephenwf, line 50: I think this is a better way of handling missing manifests, as we can detect it. If we have the error handling in the loaders, we might be able to throw an error here instead - although I'm not sure if that will cause problems elsewhere.

services/madoc-ts/src/routes/site/site-canvas.ts
```diff
@@ -15,20 +15,43 @@ export const siteCanvas: RouteMiddleware<{ slug: string; id: string }> = async c
...
+        if (plaintext && plaintext.found) {
+          context.response.body.plaintext = plaintext.transcription;
+        }
+      } catch (err: any) {
+        if (err.status === 404) {
+          context.response.body = backUpBody;
+          context.response.status = 200;
+        }
+        console.log(err);
+      }
```
stephenwf, line 47: I don't think we need this extra try/catch as it will get caught by the one further down.

stephenwf (approved):

PR #733, opened 2023-05-24
trimmed strings to 3000 character limit when importing manifest

Heather0K:


services/madoc-ts/src/gateway/tasks/import-manifest.ts
```diff
@@ -90,6 +90,25 @@ export const jobHandler = async (name: string, taskId: string, api: ApiClient) =
...
+
+      const trimIntString = (intString: InternationalString) => {
+        for (const [key, value] of Object.entries(intString)) {
+          if (value) {
+            value[0] = value[0].substring(0, 3000);
+            return intString;
+          }
+        }
+        return intString;
+      };
```
stephenwf, line 106: We might want to do this for collections/canvases later, might be worth splitting out to the helper folder

stephenwf (approved):

PR #745, opened 2023-06-06
Mad 232/feature/collecton manifest import

Heather0K:


services/madoc-ts/src/gateway/tasks/import-collection.ts
```diff
@@ -47,6 +53,7 @@ export function createTask(collectionUrl: string, userId: number, siteId?: numbe
     ],
     status: 0,
     status_text: status[0],
+    manifestIds: manifestIds,
```
stephenwf, line 56: Ah this isn't valid in the main task body, it would have to be added to the parameters (which is like the input/arguments of a task).

https://github.com/digirati-co-uk/madoc-platform/blob/ddc94d0e90ebf4e2d65ba9011dba5562b83e622b/services/madoc-ts/src/gateway/tasks/import-collection.ts#L28

A third option here with the manifest ids should work (and then update to read from the parameters instead)

For example here: https://github.com/digirati-co-uk/madoc-platform/blob/ddc94d0e90ebf4e2d65ba9011dba5562b83e622b/services/madoc-ts/src/gateway/tasks/import-collection.ts#L70

stephenwf (approved):

PR #750, opened 2023-06-08
MAD-1365 - Added conversion for config versions

stephenwf:
Adds v1 -> v2 and also v2 -> v1 for the config.

At the moment it doesn't allow new options - so not yet ready to merge.


Heather0K (approved):

stephenwf:
See #752

PR #757, opened 2023-06-19
delete all tasks when removing all contributions

Heather0K:


services/madoc-ts/src/frontend/site/features/admin/CanvasTaskProgress.tsx
```diff
@@ -84,6 +84,7 @@ export const CanvasTaskProgress: React.FC = () => {
 
   const [deleteCaptureModel, deleteCaptureModelStatus] = useMutation(async () => {
     if (canvasModel && canvasModel.model && canvasModel.model.id) {
+      await api.batchDeleteTasks({ resourceId: canvasId, subject: canvasTask?.subject });
```
stephenwf, line 87: This might delete tasks outside of the current project at the same time. The project should contain a `root_task_id` that can be used (can't remember the exact API)

services/madoc-ts/translations/en/madoc.json
```diff
@@ -971,6 +972,7 @@
   "paused": "paused",
   "pending": "pending",
   "preview": "preview",
+  "project 1": "project 1",
```
stephenwf, line 975:
```suggestion
```

PR #773, opened 2023-07-31
stacked nested and multiple autocomplete fields so its easier to read

Heather0K:


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
stephenwf, line 2: can be removed?

stephenwf (approved):

PR #779, opened 2023-08-04
WIP: saving and retreiving from locaforage

Heather0K:


services/madoc-ts/src/frontend/shared/capture-models/new/features/AutosaveRevision.tsx
```diff
@@ -1,49 +1,116 @@
...
+  if (rev.data) {
+    return (
+      <InfoMessage>
+        <TextButton
+          $inherit
+          onClick={() => {
+            setRetrievedRevision();
+          }}
+        >
+          Continue where you left off?
```
stephenwf, line 110: Needs the `t()`, can see it's already in the translation file though

PR #781, opened 2023-08-07
fix metdata dissapearing on relaod sometimes

Heather0K:


services/madoc-ts/src/frontend/site/features/CanvasMetadata.tsx
```diff
@@ -17,7 +17,7 @@ export const CanvasMetadata: React.FC<{ compact?: boolean; showEmptyMessage?: bo
   const createLink = useRelativeLinks();
   const { canvas } = useMetadataSuggestionConfiguration();
 
-  if (!data || !metadataConfig) {
```
stephenwf, line 20: There is an "isLoading" on the metadata config that we could use

stephenwf (approved):

PR #837, opened 2024-11-26
Cache project

stephenwf:
This adds 3 features:

- Disables server-side rendering on the 2 review/task pages.
- Adds a Stale-while-revalidate memory cache and uses it for slow project queries
- Adds a new request trace and in-memory store of the slowest requests. (avg, max, count)

The request trace is visible on the System page in the admin.
image

It can be disabled with the environment variable flag: `DISABLE_SLOW_REQUEST_TRACKING=true`

### DIFF
--- a/services/madoc-ts/src/utility/cache-helper.ts
+++ b/services/madoc-ts/src/utility/cache-helper.ts
@@ -1,0 +1,52 @@
+import cache from 'memory-cache';
+
+export async function cachePromise<T extends object>(
+  key: string,
+  getter: () => Promise<T>,
+  timeInMs: number
+): Promise<T> {
+  const resource = cache.get(key);
+  if (resource) {
+    return resource as T;
+  }
+
+  const result = await getter();
+  cache.put(key, result, timeInMs);
+  return result;
+}
+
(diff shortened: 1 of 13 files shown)


services/madoc-ts/src/utility/cache-helper.ts
```diff
@@ -0,0 +1,52 @@
...
+    await swcPromises[key];
+  }
+  const resource = cache.get(key);
+  const staleResource = cache.get(`@stale/${key}`);
+  if (!resource) {
+    const promise = getter();
+    swcPromises[key] = promise;
+    promise.then(result => {
+      cache.put(key, result, timeInMs);
+      cache.put(`@stale/${key}`, result, staleTimeInMs);
```
stephenwf, line 37: I checked the source code of `memory-cache`. It doesn't copy the `result`, so the extra memory to keep a stale copy is only the size of the key. The memory is released when you put the stale copy. It _will_ mean that for infrequent requests we will hold onto stale data for longer. I've set it to 8 hours for now. (The service will also restart the node process if max memory is reached).

PR #873, opened 2026-04-16
Feature/non public sites

Heather0K:


services/madoc-ts/src/frontend/site/pages/loaders/root-loader.tsx
```diff
@@ -26,6 +26,10 @@ export type RootLoaderType = {
   context: any;
 };
 
+function isAccountPagePath(pathname: string) {
+  return /\/(login|register|forgot-password|reset-password|activate-account)$/.test(pathname);
+}
```
stephenwf, line 31: We _might_ need to check the full path (as in use ^ to make sure its starts with the path) as users can define their own site pages and it could get caught. `/s/site/my-page/register` I think would get caught by this regex. I don't know if the input `pathname` starts with `/s/site/path` or just `/path`

services/madoc-ts/src/frontend/site/pages/loaders/root-loader.tsx
```diff
@@ -116,16 +120,17 @@ export const RootLoader: UniversalComponent<RootLoaderType> = createUniversalCom
...
-      },
-    ],
-    getKey: () => {
-      return ['root-config', []];
+    getKey: (params, query, pathname) => {
+      return ['root-config', { account: isAccountPagePath(pathname) }];
     },
-    getData: async (key, vars, api) => {
+    getData: async (key, vars, api, pathname) => {
+      if (isAccountPagePath(pathname)) {
```
stephenwf, line 127: I think `vars.account` would be set, to avoid the double check.

services/madoc-ts/src/middleware/site-state.ts
```diff
@@ -2,21 +2,54 @@ import { RouteMiddleware } from '../types/route-middleware';
...
+  const slug = context.params?.slug;
+  if (typeof slug === 'undefined') {
+    await next();
+    return;
+  }
+
+  const userId = context.state.jwt ? context.state.jwt.user.id : undefined;
+  const isAdmin = context.state.jwt ? context.state.jwt.scope.indexOf('site.admin') !== -1 : false;
+  const site = allowPrivateSiteAccess
+    ? await context.siteManager.getSiteBySlug(slug)
```
stephenwf, line 29: When is "allowPrivateSiteAccess" set to true (I will come back and resolve this - just as a reminder)

services/madoc-ts/translations/en/madoc.json
```diff
@@ -1303,4 +1304,6 @@
   "{{feature}} is an experimental feature": "{{feature}} is an experimental feature",
   "{{page}} of {{count}}": "{{page}} of {{count}}",
   "{{page}} of {{count}}_plural": "{{page}} of {{count}}"
+}",
+  "{{page}} of {{count}}_plural": "{{page}} of {{count}}"
 }
```
stephenwf, line 1309: Some invalid JSON to be fixed here I think.

services/madoc-ts/src/routes/user/forgot-password.ts
```diff
@@ -62,7 +62,7 @@ export const forgotPassword: RouteMiddleware = async (context, next) => {
   }
 
   // Render react.
-  await siteState(context, async () => {
+  await siteStateAllowPrivate(context, async () => {
     await siteFrontend(context, next);
   });
```
stephenwf, line 67: Hmm, I think this would leak the private site if a user didn't have access to a site but then went to the correct url (`/s/private-site/login` or `/s/private-site/register`).

PR #875, opened 2026-04-20
add ux feedback in task page and task list for missmatch and failures

Heather0K:
Screenshot 2026-04-20 at 15 39 13

### DIFF
--- a/services/madoc-ts/src/frontend/admin/pages/content/collections/import-collection.tsx
+++ b/services/madoc-ts/src/frontend/admin/pages/content/collections/import-collection.tsx
@@ -7,6 +7,7 @@ import { usePaginatedData } from '../../../../shared/hooks/use-data';
 import { ErrorIcon } from '../../../../shared/icons/ErrorIcon';
 import { Spinner } from '../../../../shared/icons/Spinner';
 import { TickIcon } from '../../../../shared/icons/TickIcon';
+import { TriangleIcon } from '../../../../shared/icons/TriangleIcon';
 import { GridContainer, HalfGird } from '../../../../shared/layout/Grid';
 import { serverRendererFor } from '../../../../shared/plugins/external/server-renderer-for';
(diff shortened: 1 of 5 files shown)


services/madoc-ts/src/frontend/admin/pages/content/collections/import-collection.tsx
```diff
@@ -127,6 +164,12 @@ export const ImportCollection: React.FC = () => {
                     </div>
                     <SystemMetadata>
                       <SystemName>{task.name}</SystemName>
+                      <SystemDescription>
+                        {t('Imported {{imported}} of {{expected}} manifests', {
```
stephenwf, line 168: Should there be a new entry in the language files?

services/madoc-ts/src/frontend/admin/pages/content/collections/import-collection.tsx
```diff
@@ -161,10 +204,39 @@ export const ImportCollection: React.FC = () => {
...
+          return task as ImportCollectionTaskWithStats;
+        }
+
+        try {
+          const stats = await api.getTaskStats(task.id, { type: 'madoc-manifest-import' });
+          const manifestImportStats: TaskManifestImportStats = {
+            expectedManifestTotal: stats.total || 0,
+            importedManifestTotal: stats.statuses?.['3'] || 0,
+            manifestStatusCounts: stats.statuses || {},
+          };
```
stephenwf, line 225: ~This seems to be running the statistics for each of the sub tasks, but I _think_ the comparison is the number of Manifests inside of the Collection. Potentially this will kick off up to 100 requests for the statistics. We _should_ be able to run a single getTaskStats() on the top level.~

~Also because of the paging these statistics only apply to the current page.~

I see now, this is the listing page for the imported items. That's totally fine!

stephenwf (approved):

PR #883, opened 2026-05-08
Fix themed header title

Heather0K:


Heather0K:
Couldn't find any dependency issues, I hid the default site title if a theme header is active

I also put an error in site configuration to help debug Ghents saving regression issue

stephenwf (approved):

